PR #125, opened 2020-05-20
Groupings-499

jayryanj:
Reviewed and modified the controller advice.


fduckart:
Hey, can we get rid of the temporary variable in the ErrorControllerAdvice#handleAccessDeniedException method?

PR #173, opened 2020-11-24
GROUPINGS-694-Create a class for global variables

zakgilbert:
[GROUPINGS-694](https://www.hawaii.edu/jira/browse/GROUPINGS-694)

### DIFF
--- a/src/main/java/edu/hawaii/its/api/access/AuthorizationServiceImpl.java
+++ b/src/main/java/edu/hawaii/its/api/access/AuthorizationServiceImpl.java
@@ -11,12 +11,8 @@ import org.springframework.stereotype.Service;
 @Service
 public class AuthorizationServiceImpl implements AuthorizationService {
 
-    @Autowired
-    private MemberAttributeService memberAttributeService;
-
-    @Autowired
-    private GroupingAssignmentService groupingAssignmentService;
-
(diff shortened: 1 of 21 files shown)


zakgilbert:
Not all the values from the properties are implemented. This pull request is to make sure that formating and implementation are acceptable before finishing the class.

fduckart (commented):
Looks like a bunch of other updates are in here for some reason. Not sure that's a good idea.

zakgilbert:
I'm just removing a few unused variables and stuff. These classes are a mess. Or do you mean the commits at the top?

fduckart:
Pull request shows 23 files updated. Unless I'm reading that incorrectly

zakgilbert:
There's a lot of files that use the @Value annotation to pull data from the properties files, doesn't that mean that all those files need to be refactored to use GrouperConfiguration instead?

fduckart:
Sure looks there's other, non-related stuff in this pull, too. Also, it might be safer if you started off with submitting just the configuration class and its related test class. Just the two classes.

zakgilbert:
If you are talking about the first few commits, those showed up cause I mistakenly branched off from my previous ticket. I ran a rebase with master and everything seems to be in order so far.

Kobeyarai808:
Outdated and would need major refactoring to implement.

PR #174, opened 2020-12-01
Dev john 548

johnbigornia:
Deleted two test cases: testAdminUser on both junit and integration, as well as any employee related test cases. Added test case in UserBuilder to test out the make function.


fduckart (commented):
Hey, can you take another crack at this one? I don't think this the right approach, imho. I don't think it's a good idea to delete the tests that were there before, but instead, we should get them to work instead of marking them with Ignore. Where things went bad before is that somebody inexplicably put the @ActiveProfiles annotation on the test, causing it to be a integration test. Our (somewhat loose) project guidelines call for a differently named test for system/integration tests. Sorry, I should I have spelled this out clearer before now.

PR #198, opened 2021-03-19
GROUPINGS-576 Fixed issue regarding threading exception handling.

jayryanj:
Modified exception handling code for threading to fix a bug where the wrong response code was being sent.


Kobeyarai808:
Old PR

PR #203, opened 2021-05-15
Added hidden property to the syncDest class so that the UI and API ca…

nzenger:
…n handle it without error.


fduckart:
Hey, would be possible to initialize that hidden variable as false right off the bat? And then also add a ternary operator in the setter so that a null is not possible.

PR #210, opened 2021-07-28
Add a simple HTML page to the '/' endpoint

chakhon:
[Grouping-756](https://www.hawaii.edu/jira/browse/GROUPINGS-756)

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/HomeController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/HomeController.java
@@ -1,0 +1,24 @@
+package edu.hawaii.its.api.controller;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.stereotype.Controller;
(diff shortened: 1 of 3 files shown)


fduckart:
Great. 

One more thing... I forgot to mention it the other day, but it would be grand if you created a test of your new controller method. See this test class from another project for an example: 
https://github.com/fduckart/uh-holiday-webapp-boot/blob/master/src/test/java/edu/hawaii/its/holiday/controller/HomeControllerTest.java

fduckart:
P.S. Slack me if you have any questions or run into a snag.

chakhon:
> 
> 
> Great.
> 
> One more thing... I forgot to mention it the other day, but it would be grand if you created a test of your new controller method. See this test class from another project for an example:
> https://github.com/fduckart/uh-holiday-webapp-boot/blob/master/src/test/java/edu/hawaii/its/holiday/controller/HomeControllerTest.java

Got it.

PR #219, opened 2021-08-14
getMemberships API call in 2.1.1 not returning properly-GROUPINGS-865

zakgilbert:
[GROUPINGS-865](https://www.hawaii.edu/jira/browse/GROUPINGS-865)

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
@@ -1,5 +1,7 @@
 package edu.hawaii.its.api.controller;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import edu.hawaii.its.api.service.GroupAttributeService;
 import edu.hawaii.its.api.service.GroupingAssignmentService;
 import edu.hawaii.its.api.service.GroupingFactoryService;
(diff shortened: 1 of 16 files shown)


mhodgesatuh:
Overall looks ok IFAICT, but some of the comments are not up to our standards, such as:

mhodgesatuh:
Overall looks ok IFAICT, but some of the comments are not up to our standards, such as:

        //get rid of duplicates
vs
        // Get rid of duplicates.


If we need to keep commented out code, should provide inline comments to explain why, because we generally don't:
        //        for (int i = 0; i < 6; i++) {
        //            membershipService.deleteGroupMemberByUsername(ownerUsername, GROUPING_INCLUDE, username[i]);
        //        }

fduckart:
Those two new tests in the GroupingAssignmentServiceTest class... I'm not exactly sure what the bottom part of each one is suppose to prove.

fduckart:
In MembershipServiceTest#getMembershipResultsTest ... that try/catch exception test needs a fail call in the middle so you can make sure the error is happening.

zakgilbert:
For your first comment you are talking about these right 
@Test
    public void getOptOutGroupsTest() {
        List<String> optOutPaths =
                groupingAssignmentService.getOptOutGroups(users.get(0).getUsername(), users.get(1).getUsername());
        assertNotNull(optOutPaths);
        Set<String> pathMap = new HashSet<>();
        for (String path : optOutPaths) {
            assertTrue(pathMap.add(path));
        }
    }

    @Test
    public void getOptInGroupsTest() {
        List<String> optInPaths =
                groupingAssignmentService.getOptInGroups(users.get(0).getUsername(), users.get(1).getUsername());
        assertNotNull(optInPaths);
        Set<String> pathMap = new HashSet<>();
        for (String path : optInPaths) {
            assertTrue(pathMap.add(path));
        }
    }

PR #220, opened 2021-08-17
getNumberOfMemberships method and tests

chakhon:
[GROUPINGS-839](https://www.hawaii.edu/jira/browse/GROUPINGS-839)


fduckart:
Yo, re: getNumberOfMembershipTest... it passes, but it's not actually doing what you want, I don't think.

PR #223, opened 2021-09-02
/groupings - implement new 'hidden' attribute to determine which Sync destination can be displayed-GROUPINGS-768

zakgilbert:
[GROUPINGS-768](https://www.hawaii.edu/jira/browse/GROUPINGS-768)


fduckart:
I kind of think you should keep the line you took out of GroupAttributeServiceImpl. 

The SyncDestination class, I think you should make some changes. 

The SyncDestinationTest class, maybe we can clean that up, too.

One research item: find out how many references we have to the non-parameter constructor of SyncDestination.

fduckart (commented):
In the constructor, use the setter for the tooltip like the other cases.

PR #224, opened 2021-09-07
Groupings-854 

ruizaj:
Adding assertThat() tests into each function within rest controller test file to check if that fixes Codacy errors.


mhodgesatuh:
Looks fine to me.  I notice that the changes are also designed to appease codacity.

Kobeyarai808:
This looks good. Looks like you took Franks advice!

PR #227, opened 2021-09-13
Adding opencsv v5.0 dependencies

ruizaj:


fduckart:
I don't get it. Just a pom update? Is/was this part of another commit?

ruizaj:
It was a task for ticket 896. Just a small pom commit.

fduckart:
I guess I'm just confused by the web interface of github or something. All I see is the addition of 5.0 opencsv, but I don't see the removal of the old duplicate dependency? That's what I don't get.

mhodgesatuh:
For the master branch the 3.5 entry needs to be removed.  Andre explains the 2.1.1 branch there are no opencvs entries at all, so it is being added for the first time.

fduckart:
Thanks, I get it now.

fduckart:
Was thinking about this while running around Diamond Head... this update just can't be right.

PR #236, opened 2021-10-12
Groupings-906

ruizaj:
This also includes my changes of Groupings-815 without the tests since that is being worked on by Ryota and Luke


mhodgesatuh:
What are the plans for back porting some of the above to dev-2-2-0?

PR #243, opened 2021-10-30
Fix Remove privileges

yiwenc22:
Before only owner can remove member from groups, now it will let admin do it too.


fduckart:
The lines in those updated if-statements are getting kind of long, I think. Maybe it would be good to create a new private method called, I don't know, maybe like "isAccessPermitted(groupingPath, currentUser)" and update all the places necessary. Or maybe even the opposite naming (e.g., "isAccessDenied"). Whichever. Also, these kind of changes would usually come with test file updates, too.

yiwenc22:
> The lines in those updated if-statements are getting kind of long, I think. Maybe it would be good to create a new private method called, I don't know, maybe like "isAccessPermitted(groupingPath, currentUser)" and update all the places necessary. Or maybe even the opposite naming (e.g., "isAccessDenied"). Whichever. Also, these kind of changes would usually come with test file updates, too.

yiwenc22:
> The lines in those updated if-statements are getting kind of long, I think. Maybe it would be good to create a new private method called, I don't know, maybe like "isAccessPermitted(groupingPath, currentUser)" and update all the places necessary. Or maybe even the opposite naming (e.g., "isAccessDenied"). Whichever. Also, these kind of changes would usually come with test file updates, too.

Ok, will do that and create a new pull request. Thank for the suggestion

PR #247, opened 2021-11-17
Export to CSV Button in MultiAdd members modal doesn't work.

zakgilbert:
Check for null uid in remove and add member functions, if null then retrieve.
[GROUPINGS-922](https://www.hawaii.edu/jira/browse/GROUPINGS-922).


zakgilbert:
Note that this is not a completed ticket, but these changes are needed asap. Another pull request will be open once the modal is fixed.

mhodgesatuh:
Looks good

mhodgesatuh:
And we need to backport the fix to 2.2 I assume.

ruizaj:
I'll commit this one in, looks fine to me.

PR #251, opened 2021-12-04
ticket 871

andywritescode:


mhodgesatuh:
//swaps the owner out with the admin
is improved by
        // Makes the admin also the owner in the event that there are no remaining owners otherwise.

Notice the use of proper English sentence structure for our inline comments.

ruizaj:
Just added the comment changes, but otherwise this looks good to me.

PR #252, opened 2021-12-10
Groupings-941: Fixed all tests except memberAttributesAdminTest()

ruizaj:
Will be creating a ticket to address this final test.

### DIFF
--- a/src/test/java/edu/hawaii/its/api/service/TestGroupAttributeService.java
+++ b/src/test/java/edu/hawaii/its/api/service/TestGroupAttributeService.java
@@ -203,7 +203,7 @@ public class TestGroupAttributeService {
         assertTrue(memberAttributeService.isOwner(GROUPING, username[0]));
         assertTrue(groupAttributeService.isGroupAttribute(GROUPING, LISTSERV));
 
-        //get last modified time
+        // Get last modified time
         WsGetAttributeAssignmentsResults attributes =
                 groupAttributeService.attributeAssignmentsResults(ASSIGN_TYPE_GROUP, GROUPING, YYYYMMDDTHHMM);
         String lastModTime = attributes.getWsAttributeAssigns()[0].getWsAttributeAssignValues()[0].getValueSystem();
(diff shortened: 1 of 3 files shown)


zakgilbert:
This is a 2.2 ticket so a separate branch needs to be made off of dev-2-2-0.

src/test/java/edu/hawaii/its/api/service/TestGroupAttributeService.java
```diff
@@ -247,17 +247,17 @@ public void changeListServStatusTest() {
         String lastModTime4 = attributes.getWsAttributeAssigns()[0].getWsAttributeAssignValues()[0].getValueSystem();
         assertThat(lastModTime4, is(lastModTime3));
 
-        assertFalse(memberAttributeService.isOwner(GROUPING, username[1]));
+        // Invalid user can't change attribute status
         try {
-            groupAttributeService.changeGroupAttributeStatus(GROUPING, username[1], LISTSERV, true);
+            groupAttributeService.changeGroupAttributeStatus(GROUPING, "zzzzz", LISTSERV, true);
```
zakgilbert, line 252: Is "zzzzz" a possible uhUsername?

fduckart:
Shouldn't there be "fail" calls in those tests in the enableDisablePreferencesFailTest method?

mhodgesatuh:
Question: is 'zzzzzz' potentially a valid uhUuid (UH Number)?  Unlikely, but, technically, yes.  Place and underscore in the middle like this" 'zzz_zzz' and you have picked a username that is not in our namespace given our formatting rules.

src/test/java/edu/hawaii/its/api/service/TestGroupAttributeService.java
mhodgesatuh: Question: is 'zzzzzz' potentially a valid uhUuid (UH Number)? Unlikely, but, technically, yes. Place and underscore in the middle like this" 'zzz_zzz' and you have picked a username that is not in our namespace given our formatting rules.

PR #253, opened 2021-12-21
Add multi-remove feature to owners list #901

devgav:


Kobeyarai808:
This looks good to me, with no regressions. If you are satisfied with this, feel free to push.

Kobeyarai808 (commented):
Looks good as stated before.

zakgilbert (approved):

Kobeyarai808 (approved):

fduckart:
It would be grand if you could fix up some of the existing problems in the test Java classes you updated... especially TestMemberAttributeService.java.

devgav:
> It would be grand if you could fix up some of the previously existing problems in the test Java classes you updated... especially TestMemberAttributeService.java.

 Right now I am not sure how to fix the problems in TestMemberAttributeService.java. But I can create a ticket so that someone can look into it. I am eager to get these changes out so that I can finish [GROUPINGS-696](https://www.hawaii.edu/jira/browse/GROUPINGS-696) since this update will have the assignOwnership moved to the membershipService.java and also some other important refractoring.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
```diff
@@ -335,20 +335,20 @@ public void init() {
...
-                .body(memberAttributeService.assignOwnership(path, currentUser, uid));
+                .body(membershipService.assignOwnership(path, currentUser, uid));
     }
 
     /**
      * Delete a grouping owner.
      */
-    @DeleteMapping(value = "/groupings/{path:[\\w-:.]+}/owners/{uid:[\\w-:.]+}")
-    public ResponseEntity<GroupingsServiceResult> deleteOwner(@RequestHeader("current_user") String currentUser,
+    @DeleteMapping(value = "/groupings/{path:[\\w-:.]+}/removeOwners/{uid}")
```
zakgilbert, line 344: The path can be changed to just /owners/ instead of /removeOwners/, because that is the RESTfull format and the HTML DELETE function already implies removal. You might want to make {uid} plural to indicate that multiple owners can be removed as well.

src/test/java/edu/hawaii/its/api/service/TestMemberAttributeService.java
```diff
@@ -495,4 +498,13 @@ public void getOwnedGroupingsTest() {
             e.printStackTrace();
         }
     }
-}
+
+    @Test
+    public void multiRemoveOwners() {
```
zakgilbert, line 503: Could change this to removeOwnershipsTest to follow the format we have been using.
<nameOfMethodBeingCovered>Test

PR #255, opened 2021-12-29
fix memberAttributesAdminTest()

chakhon:
[GROUPINGS-951](https://www.hawaii.edu/jira/browse/GROUPINGS-951)

- Main problem was the attributes defined in the application.properties file does not match the attributes within the person object that we were returning in getMemberAttributes().
- Another problem was all test users (i.e. iamtst01) was a owner of some group, so it was failing the Current user is not admin or owner test.
- Created isAdminTest() and isOwnerTest() for troubleshooting
- Created [GROUPINGS-961](https://www.hawaii.edu/jira/browse/GROUPINGS-961)

Let me know if theses changes need to be ported to dev-2-2-0


chakhon:
Seems Zach has address the problem in a better way in #258, and it has already been merged,

PR #256, opened 2021-12-30
Add multi-add feature to owners #696

devgav:
[GROUPINGS-696](https://www.hawaii.edu/jira/browse/GROUPINGS-696)

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/MembershipService.java
+++ b/src/main/java/edu/hawaii/its/api/service/MembershipService.java
@@ -10,6 +10,10 @@ import java.util.List;
 public interface MembershipService {
 
     boolean isUhUuid(String username);
+
+    GroupingsServiceResult assignOwnership(String groupingPath, String ownerUsername, String newOwnerUsername);
+
+    List<RemoveMemberResult> removeOwnerships(String groupingPath, String actor, List<String> ownersToRemove);
    
     List<Membership> getMembershipResults(String owner, String uid);
 
--- a/src/main/java/edu/hawaii/its/api/service/MembershipService.java
+++ b/src/main/java/edu/hawaii/its/api/service/MembershipService.java
@@ -2,20 +2,15 @@ package edu.hawaii.its.api.service;
 
 import edu.hawaii.its.api.type.AddMemberResult;
 import edu.hawaii.its.api.type.GroupingsServiceResult;
-import edu.hawaii.its.api.type.Membership;
 import edu.hawaii.its.api.type.RemoveMemberResult;
 
 import java.util.List;
 
 public interface MembershipService {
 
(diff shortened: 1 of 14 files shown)


src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -11,7 +11,7 @@
 
     boolean isUhUuid(String username);
 
-    GroupingsServiceResult assignOwnership(String groupingPath, String ownerUsername, String newOwnerUsername);
+    List<AddMemberResult> assignOwnerships(String groupingPath, String ownerUsername, List<String> newOwnerUsername);
```
zakgilbert, line 14: @mhodgesatuh I'm thinking we should rename it to `addOwners`?

Kobeyarai808: I really like _assignOwnership(s)_ because the meaning is clearer in my opinion. 

I feel like we are assigning the role owner to the user. If we we are to use _addOwner_, I feel like it is not clear what we are adding the owner to. 

_AddOwners_ would follow our naming convention better though. 

Thoughts? I think either way would work just as well. @zakgilbert @mhodgesatuh @devgav @fduckart

devgav: how about addOwnerships? only because in the uh-groupings-ui we have a $scope.addOwners() and it may be confusing.

zakgilbert: Should the naming be consistent all the way through @mhodgesatuh?

devgav:
I am going to split the add and remove ownership function and probably combine the multi add and multi remove into the new add and remove ownership functions

mhodgesatuh:
Codacy is flagging some unused imports.  Your IDE should also be able to flag them:
  o Unused import - edu.hawaii.its.api.type.Membership.
  o Unused import - edu.hawaii.its.api.repository.GroupingRepository.

fduckart:
I can also go over how we can modify the pom to include a plugin that will give us the capability to run this kind of check locally.

devgav:
Splitting the getMembershipResults into a smaller function seems to have gotten rid of the codacy issue that I was experiencing, I plan to write a test code for the new function I created so that we don't decrease our code coverage.

src/main/java/edu/hawaii/its/api/service/MembershipService.java
devgav: after the meeting I changed the name to addOwners

zakgilbert:
I'll look at it when I get home

Kobeyarai808:
This is a old ticket.

PR #258, opened 2022-01-04
fix enableDisablePreferencesFailTest()

zakgilbert:
Pass a user that will fail.
Fix Objectmapper error.
[GROUPINGS-961](https://www.hawaii.edu/jira/browse/GROUPINGS-961)

### DIFF
--- a/src/test/java/edu/hawaii/its/api/controller/TestGroupingsRestControllerv2_1.java
+++ b/src/test/java/edu/hawaii/its/api/controller/TestGroupingsRestControllerv2_1.java
@@ -23,6 +23,7 @@ import edu.hawaii.its.api.type.Group;
 import edu.hawaii.its.api.type.Grouping;
 import edu.hawaii.its.api.type.GroupingsHTTPException;
 import edu.hawaii.its.api.type.GroupingsServiceResult;
+import edu.hawaii.its.api.type.Person;
 
 import edu.internet2.middleware.grouperClient.ws.beans.WsSubjectLookup;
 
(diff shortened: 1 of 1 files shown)


src/test/java/edu/hawaii/its/api/controller/TestGroupingsRestControllerv2_1.java
```diff
@@ -270,6 +272,7 @@ public void setUp() throws IOException, MessagingException {
         uhUser02 = new User(usernames[1], usernames[1], uhAuthorities);
         uhUser05 = new User(usernames[4], usernames[4], uhAuthorities);
         uhUser03 = new User(usernames[2], usernames[2], uhAuthorities);
+        bogusUser = new User("bogus-user", "bogus-user", uhAuthorities);
```
zakgilbert, line 275: If we are testing for a failure, then create a bogus user.

src/test/java/edu/hawaii/its/api/controller/TestGroupingsRestControllerv2_1.java
```diff
@@ -375,11 +378,12 @@ public void memberAttributesAdminTest() throws Exception {
...
         // Current user is admin.
         MvcResult result = mockMvc.perform(get(API_BASE + "members/" + usernames[0])
-                .header(CURRENT_USER, ADMIN)
-                .with(user(ADMIN))
-                .with(csrf()))
+                        .header(CURRENT_USER, ADMIN)
+                        .with(user(ADMIN))
+                        .with(csrf()))
                 .andReturn();
-        Map map = new ObjectMapper().readValue(result.getResponse().getContentAsByteArray(), Map.class);
```
zakgilbert, line 382: We were mapping a person object to a map, which was breaking.

src/test/java/edu/hawaii/its/api/controller/TestGroupingsRestControllerv2_1.java
```diff
@@ -822,14 +829,14 @@ public void enableDisablePreferencesAndSyncDestsPassTest() throws Exception {
     @Test
     public void enableDisablePreferencesFailTest() throws Exception {
         try {
-            mapGSRs(API_BASE + "groupings/" + GROUPING + "/preferences/" + OPT_IN + "/disable", "put", uhUser02);
+            mapGSRs(API_BASE + "groupings/" + GROUPING + "/preferences/" + OPT_IN + "/disable", "put", bogusUser);
             fail("Shouldn't be here.");
```
Kobeyarai808, line 826: Just a suggestion, we can write why the test should not reach this fail.

Kobeyarai808 (approved):
Overall, very clean. I like these changes.

src/test/java/edu/hawaii/its/api/controller/TestGroupingsRestControllerv2_1.java
zakgilbert: Wow-what a careless mistake.

mhodgesatuh:
Might be useful to include inline documentation explaining that the purpose of putting a hyphen in the middle of the bogus username is to ensure that the bogus username doesn't accidentally match a real username since real usernames can't have a hyphen in the middle.

PR #270, opened 2022-01-15
Fix JUnit tests

devgav:
- [x] Unit tests
- [x] Integration tests


mhodgesatuh:
membershipService.removeOwnerships(GROUPING, usernames[0], Arrays.asList(usernames[1]));

Frank lead a good discussion at yesterday's meeting about why that third parameter doesn't smell right and what to do about it via an overloaded method declaration.

zakgilbert:
> ```
>     membershipService.removeOwnerships(GROUPING, usernames[0], Arrays.asList(usernames[1]));
> ```
> 
> Frank lead a good discussion at yesterday's meeting about why that third parameter doesn't smell right and what to do about it via an overloaded method declaration.

Wasn't frank referring to using Collections.singleton?

mhodgesatuh:
Perhaps, but to my eye this seems to fit the same pattern.  It is certainly possible that I'm not seeing this in the correct context.  Frank has been added as another reviewer.

zakgilbert:
> Perhaps, but to my eye this seems to fit the same pattern. It is certainly possible that I'm not seeing this in the correct context. Frank has been added as another reviewer.

Was there a ticket made to go through the code and spray some Febreze on these overloaded declarations? They are all over the code base.

zakgilbert (approved):

mhodgesatuh:
As Branch Lead, you can make the call and let this through.  Given that these are everywhere, we need a ticket to ensure we address this later.  Thanks.

Kobeyarai808:
My thoughts: 
At this point, I would consider this a large refactor of the tests. I am very worried that we will be removing critical tests and lower our code coverage. We did not make large edits to the tests in a while and I am not sure why removing them now makes the test pass. I am willing to merge this if everyone is on the same page since I don't know the tests that well.

Kobeyarai808 (approved):
If everyone is on the same page, I am for this merge.

mhodgesatuh:
aok to merge

PR #274, opened 2022-01-19
Dependabot PR: Bump h2 from 1.4.198 to 2.0.206 

zakgilbert:
Move back to the previous h2 version, the move up may have been the cause of the stack breakage today.
[GROUPINGS-989](https://www.hawaii.edu/jira/browse/GROUPINGS-989)


zakgilbert:
Pushing this to return the pom back. If Jenkins passes tomorrow morning then we know we need to look into this PR more.

PR #288, opened 2022-03-04
Fix memberships showing incorrect number

devgav:
[GROUPINGS-1042](https://www.hawaii.edu/jira/browse/GROUPINGS-1042)
- [x] Unit Test 
- [x] Integration Test

It will now not include the excludes in the membership count we display on the home page only the basis and the include count.

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/MembershipServiceImpl.java
+++ b/src/main/java/edu/hawaii/its/api/service/MembershipServiceImpl.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 @Service("membershipService")
 public class MembershipServiceImpl implements MembershipService {
@@ -200,7 +201,6 @@ public class MembershipServiceImpl implements MembershipService {
     @Override public List<Membership> getMembershipResults(String owner, String uid) {
         String action = "getMembershipResults; owner: " + owner + "; uid: " + uid + ";";
         logger.info(action);
-
         if (!memberAttributeService.isAdmin(owner) && !owner.equals(uid)) {
             throw new AccessDeniedException(INSUFFICIENT_PRIVILEGES);
         }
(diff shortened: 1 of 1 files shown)


Kobeyarai808 (approved):
Looks good in my eyes. I inquired why this ticket was necessary with Gavin given the discussion in today's meeting. He replied with 

"yeah, basically I went back and relooked at the getMembershipsResult function and saw that the function was just adding all the groupings that they were a member of (basis, include, exclude) but we don't count the exclude because they are not considered a member of a group (edited) "

I think this was a really good catch!

fduckart:
Hey, can you fix the formatting on getNumberOfMemberships? Also, what's with that int cast there?

devgav:
> Hey, can you fix the formatting on getNumberOfMemberships? Also, what's with that int cast there?

The .count() counts the number of elements in the stream and return a long, but the return type is an int so I decided to typecast it, idk whether this is good practice to type cast, if it isn't here is another way to write it which will return an int.

`// Get the number of memberships the current user has
    @Override public Integer getNumberOfMemberships(String currentUser, String uid) {
        return getMembershipResults(currentUser, uid)
                .stream()
                .filter(membership -> membership.isInInclude() || membership.isInBasis())
                .collect(Collectors.toList())
                .size();
    }`

fduckart:
Yeah, I kinda got the stream count as long issue, I guess I was wondering if we should dig all the way down to the base call and make that return a Long, so then we wouldn't run into the mismatch.

fduckart:
Also, I forgot to mention/comment on this: Is there a call that returns the 'membership.isInInclude() || membership.isInBasis()' filtering directly? It kind of appears we are getting more data that we need via the call, then filtering out what we don't want. I guess that's perfectly fine if most of data is those two cases, but if not, it seems a bit inefficient. Just a thought, though, and I'm not sure it's on the mark.

devgav:
> 

As far as I know there isn't a method to get only the include and in basis groups directly. The getMembershipResult only gets the list of memberships that match a users id, so it is only returning their basis, include, exclude, and owners grouping paths. Of the 4 I believe the basis and include should be the most frequent that appear most of the time, although I would need a confirmation whether this is true from @mhodgesatuh.

zakgilbert (approved):

zakgilbert (commented):
Yeah sorry @devgav  I merged this too quickly, please change the name to just 
    `numberOfMemberships`
the get is implied. This can be applied to the endpoint as well.

mhodgesatuh:
If this is for the memberships count as it appears on the home page, then for each Grouping a person is get's a count of 1 if the person is (in the basis but not exclude) or (in the include).  The person is not a member if they are (in the exclude) or if they are (not in the basis and not in the include).  Hopefully that makes sense.

PR #303, opened 2022-05-04
Update Coverage Badge Reference

Erick-Orozco-Ciprian:
# Ticket Link

[Ticket](https://www.hawaii.edu/jira/browse/GROUPINGS-1068)

# List of squashed commits

- Changed coverage status badge reference

# Test Checklist

- [x] Unit Tests Passed:


mhodgesatuh (commented):
FYI, only need to test the LTS versions of Java: 8, 11, 17 (not 18).  I note that the 11+ tests are failing.  Might be good to better understand why before proceeding since we will be upgraded to 11 or 17 in the future.

Update: I see now that those tests are in progress, so the only issue is that we should be testing 17 and not 18.

Kobeyarai808 (approved):
Looks good to me!

Erick-Orozco-Ciprian:
> FYI, only need to test the LTS versions of Java: 8, 11, 17 (not 18). I note that the 11+ tests are failing. Might be good to better understand why before proceeding since we will be upgraded to 11 or 17 in the future.
> 
> Update: I see now that those tests are in progress, so the only issue is that we should be testing 17 and not 18.

Sounds Good! I'll update it so it tests 17 not 18 with the next PR coming soon which updates the badges branch ReadMe as well.

Kobeyarai808 (approved):
Looks good!

devgav (approved):

PR #315, opened 2022-06-03
remove unnecessary activeProfile localTest

Rubegen:
# Ticket Link

[Ticket](Insert Link Here)

# List of squashed commits

- Commit 1
- Commit 2
- Commit 3

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/src/test/java/edu/hawaii/its/api/configuration/AppConfigRunTest.java
+++ b/src/test/java/edu/hawaii/its/api/configuration/AppConfigRunTest.java
@@ -7,7 +7,6 @@ import org.springframework.test.context.ActiveProfiles;
 
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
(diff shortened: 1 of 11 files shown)


mhodgesatuh:
There are no checks on the checklist above, so Gaven doesn't know if you've went through it successfully.

devgav:
We should be careful that removing these ActiveProfiles don't 

1. influence any .properties files
2. the beans loaded into the application context

i believe we use these ActiveProfiles to control the variables we have in our .properties file in src/main/resources

fduckart (commented):
Awesome!

devgav (approved):
I checked these files with the master API branch and seems like we don't use any of our variables from our .properties files. I can approve this message and go ahead and merge it thanks for  removing unnesscary code.

PR #338, opened 2022-07-28
Added descriptions to groupPath and membership classes. Added FindGroupsResults wrapper

JorWo:
# Ticket Link

[Groupings-533](https://www.hawaii.edu/jira/browse/GROUPINGS-533)

# List of squashed commits

- Changed makePaths in HelperServiceImpl to use parallel streams
- Added description to constructors, getters, and setters of groupPath and membership type classes
- Added description testing for MembershipTest and GroupingPathTest
- Changed optInPaths in GroupingAssignmentServiceImpl to use parallel streams
- Updated HelperServiceTest to include a MockBean of GrouperApiService
- Added FindGroupsResults class, Results class, Resultable interface and created tests with 100% coverage for them
- Extended SubjectResults and AttributeAssignmentResults with Results to bypass hasEmpty() testing coverage (GroupsResults is last wrapper not extended with Results)

# Test Checklist

- [x] Unit Tests Passed:


mhodgesatuh:
Does the test properties file contain any real data?  If so, that data needs to be edited to not be real before this can be published to github.  These properties files make me nervous because of the potential to post real data on the web, which would be a terrible outcome.  Real data is UH data that is assigned to a real person.

JorWo:
> Does the test properties file contain any real data? If so, that data needs to be edited to not be real before this can be published to github. These properties files make me nervous because of the potential to post real data on the web, which would be a terrible outcome. Real data is UH data that is assigned to a real person.

Well, I used the integration tests to create the data, then changed the UH username that to one that doesn't exist, named "Obiwan." The uuid 441393d72ba9421e9c0401b3c07c1650 is the only thing I'm not sure about, but it was generated from our integration tests.

mhodgesatuh:
in grouper.test.properties I also see gavin4.

Probably best to vet these files with one more person before allowing them to touch github.  Glad to hear that you are sensitive to this issue.

Longer term, I think we need to handcraft a set of data that can be assembled for testing purposes to ensure we don't leak data.

devgav:
I also see Will's data in there as well. Could you change up the data in there so that Will(wliang80) and my uh username are not shown in the grouper.test.properties file. Thank you

mhodgesatuh:
We are going to need a better vetting process for ensuring data fixtures for testing do not include real usernames/numbers.

mhodgesatuh:
Meanwhile, we need to remove the existing src/test/resources/grouper.test.properties from github history.  There’s an article on how to do that as I recall.

devgav:
I think we were going with just outright removing the grouper.test.properties file and remaking it in another ticket...

fduckart:
Wouldn't removing the file break a bunch of tests?

devgav:
> 

that is a good consideration. Then removing the usernames would def be the safest option instead of removing the file.

devgav (approved):
Seems like the issue with having real user data in our grouper.test.properties has been resolved which was my concern with this pr.

PR #347, opened 2022-08-11
Implement try...when and add comment stating scanner is closed at end of try

devgav:
# Ticket Link

[GROUPINGS-1149](https://uhawaii.atlassian.net/browse/GROUPINGS-1149)

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/PatternPropertyChecker.java
+++ b/src/main/java/edu/hawaii/its/api/service/PatternPropertyChecker.java
@@ -15,7 +15,6 @@ public class PatternPropertyChecker {
 
     private static final Log logger = LogFactory.getLog(PatternPropertyChecker.class);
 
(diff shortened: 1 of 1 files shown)


mhodgesatuh:
I thought we agreed to leave the code as is because it is working and to just add an inline comment to clarify that the try block closes the resource.  Or was that a different conversation?

devgav:
> I thought we agreed to leave the code as is because it is working and to just add an inline comment to clarify that the try block closes the resource. Or was that a different conversation?

This pr adds the inline comment inside above the try..when

mhodgesatuh (approved):
Thanks for adding the inline comment

fduckart:
This reminds of all the signs people put up in the ITC building telling us stuff that's already obvious... LOL. I still wish we could track down what is going on. I suspect there's probably a real bug somewhere, but we need to suss it out, if it's there. I'll look at the code today if I get a chance.

PR #353, opened 2022-09-01
Addresses Dependabot Alerts

JorWo:
# Ticket Link

[Groupings-1122](https://uhawaii.atlassian.net/browse/GROUPINGS-1122)

# List of squashed commits

- Bumped Bouncy Castle to 1.70
- Bumped Guava to 31.1-jre, excluded Guava from springfox-swagger2

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,11 +120,6 @@
             <artifactId>slf4j-ext</artifactId>
             </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>26.0-jre</version>
-        </dependency>
-        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
             </dependency>
@@ -145,6 +140,12 @@
             <groupId>io.springfox</groupId>
             <artifactId>springfox-swagger2</artifactId>
             <version>${springfox.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.springfox</groupId>
@@ -207,7 +208,12 @@
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcprov-jdk15on</artifactId>
-            <version>1.60</version>
+            <version>1.70</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>31.1-jre</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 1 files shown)


mhodgesatuh:
It would be good to check our Spring Boot version to see if the two dependency updates are known to be officially tested by the Spring project.  Generally, we update dependencies by starting at the top, Spring Boot being an example of the top.

JorWo:
> It would be good to check our Spring Boot version to see if the two dependency updates are known to be officially tested by the Spring project. Generally, we update dependencies by starting at the top, Spring Boot being an example of the top.

I can't seem to find any documentation on dependency compatibilities between Spring Boot with Bouncy Castle or Guava. However, being able to run the API and pass all unit and integration tests seems to be a good sign.

devgav:
I might need to defer to someone a bit more experienced with these different spring-boot dependencies. Possibly @fduckart or @zakgilbert have more input than I might about the guava and bouncy castle dependencies.

mhodgesatuh:
Did we try removing each of these and running the regression and integration tests to determine if one or both could be removed entirely from the project?  If we determine that either, or both are not needed, that would reduce our attack surface.

JorWo:
> Did we try removing each of these and running the regression and integration tests to determine if one or both could be removed entirely from the project? If we determine that either, or both are not needed, that would reduce our attack surface.

I asked Frank and found out that we have two instances of Bouncy Castle and Guava in our maven dependencies. For Bouncy Castle, one instance is installed directly in the pom.xml and another instance is installed as part of the dependency for jasig cas-client-core. The same goes for Guava in springfox swagger2. So in this PR, I excluded Bouncy Castle from jasig cas-client-core and Guava from springfox swagger2.

mhodgesatuh:
Re: I excluded Bouncy Castle from jasig cas-client-core and Guava from springfox swagger2.
Is the the course of action Frank recommended?  I'd have thought we'd remove our POM entries since we have a guiding principle to avoid specifying dependency versions in order to avoid implementing untested combinations.

JorWo:
> Re: I excluded Bouncy Castle from jasig cas-client-core and Guava from springfox swagger2. Is the the course of action Frank recommended? I'd have thought we'd remove our POM entries since we have a guiding principle to avoid specifying dependency versions in order to avoid implementing untested combinations.

I made a mistake when excluding Bouncy Castle from jasig cas-client. The artifact of Bouncy Castle in jasig cas-client is bcprix-jdk15on while the one entered in the pom is bcprov-jdk15on. It also seems that the snyk test does not pass without the Bouncy Castle bcprov-jdk15on dependency.

devgav:
Can you check and report the condition of Swagger after these changes? 

i.e. Does Swagger run and do the endpts return the appropriate data?

JorWo:
> Can you check and report the condition of Swagger after these changes?
> 
> i.e. Does Swagger run and do the endpts return the appropriate data?

All parts of Swagger is working, but I am currently looking into using Springfox 3.0.0 because that version removes their Guava dependency.

mhodgesatuh:
As discussed, we should see if they are actually needed since we don't see
them explicitly listed in the Spring docs.

PR #359, opened 2022-09-19
Fix error when trying to do an invalid multi-add in Owners

devgav:
# Ticket Link

[GROUPINGS-1182](https://uhawaii.atlassian.net/browse/GROUPINGS-1182)

# Test Checklist

- [x] Unit Tests Passed:


mhodgesatuh (approved):
I don't see any issues.

fduckart:
I understand this, but wish we weren't doing exception-based control logic.

devgav:
> I understand this, but wish we weren't doing exception-based control logic.

I think that this is part of a larger issue that I can separate into another ticket. I am sure this is intentional but the UI actually looks to check if we have any failed adds in our "handleSuccessfulAdd" and the try/catch there is there to make sure we add that failed add otherwise i think it will just throw an error.

devgav:
Created ticket regarding using exception-based control logic in our API -> https://uhawaii.atlassian.net/browse/GROUPINGS-1223

PR #384, opened 2022-11-05
Fixes getNumberOfMemberships and its test

KaileeHung:
# Ticket Link

[GROUPINGS-1262](https://uhawaii.atlassian.net/browse/GROUPINGS-1262)

# List of squashed commits

- Fixes getNumberOfMemberships and its test
- Finalizes additions to getNumberOfMembershipsTest
- Optimizes getNumberOfMemberships
- Optimizes numberOfGroupings
- Fixes missing EOL

# Test Checklist

- [X] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
@@ -465,7 +465,7 @@ public class GroupingsRestControllerv2_1 {
         logger.info("Entered REST getNumberOfGroupings...");
         return ResponseEntity
                 .ok()
-                .body(memberAttributeService.getNumberOfGroupings(currentUser, uid));
+                .body(memberAttributeService.numberOfGroupings(currentUser, uid));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


mhodgesatuh (approved):
Very nice to also see the corresponding unit test.

fduckart:
This is anal, but can you update the comment to use a leading upper-case letter and a period at the end. And can you format the comment so that it doesn't go past the 80 column marker? Just make it a two-line comment, if necessary.

I'm thinking it would be super cool if we have a custom Predicate for that filter. Not sure if it should be a static helper method or a custom class altogether, but it should have a nice name telling us what it does.

Lastly... I think it would be great if we took this further... maybe a thinner method to count the memberships. The current method creates whole structures of data, then just ignores it all just to get the count. A new, thinner method could perhaps return the integer we want it to be, too, maybe.

devgav (commented):
This looks fine just make the changes regarding Frank's comment 👍.

devgav:
Can you fix that missing EOL as well.

devgav (approved):
Looks great. I spoke with Kailee and said she was going to work on another ticket regarding Frank's suggestions.

PR #388, opened 2022-11-29
Edits membershipResults and adds new endpoint managePersonResults

KaileeHung:
# Ticket Link

Done in conjuction with: https://github.com/uhawaii-system-its-ti-iam/uh-groupings-ui/pull/679 on the UI

[GROUPINGS-1278](https://uhawaii.atlassian.net/browse/GROUPINGS-1278)

# List of squashed commits

- Refactors getNumberOfMemberships to numberOfMemberships
- Adds filter to membershipResults and creates managePersonResults
- Creates new endpoint for managePersonResults
- Updates membershipResultsTest and creates managePersonResultsTest
- Adds fixes to other tests

# Test Checklist

- [X] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
@@ -197,6 +197,19 @@ public class GroupingsRestControllerv2_1 {
     }
 
     /**
+     * Get a list of all groupings pertaining to uid (nonfiltered).
+     */
+    @GetMapping(value = "/members/{uid:[\\w-:.]+}/groupings/all")
+    @ResponseBody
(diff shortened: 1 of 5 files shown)


devgav:
This seems fine thanks for adding in the testing as well.

devgav (approved):

fduckart:
I think I liked it a little better when MembershipService#createMembershipList called that createPathMap helper method. Not because it's especially great, but since the if statements in there are a bit weird-looking, and refactoring them in the helper method will be a little easier. 

Also, I'm a bit leery of those try-catch blocks. Not sure why we have those... the called methods obviously must throw them... but it's unfortunate we have Grouper API exceptions bubbling up.

devgav:
Here is the ticket link discussing Franks changes. Seems like they have decided that this was a seperate concern.

PR #390, opened 2022-12-08
UpdateLastModified API call appears to be significantly delayed

zakgilbert:
# Ticket Link

[GROUPINGS-1295](https://uhawaii.atlassian.net/browse/GROUPINGS-1295)

# List of squashed commits
- Set last-modified:yyyymmddThhmm on grouping include and/or exclude subgroups.
- Set the timestamp one minute into the future.
- Add logger info to add/remove functions.

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/GrouperFactoryServiceImpl.java
+++ b/src/main/java/edu/hawaii/its/api/service/GrouperFactoryServiceImpl.java
@@ -1,5 +1,7 @@
 package edu.hawaii.its.api.service;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import edu.hawaii.its.api.type.Person;
 import edu.hawaii.its.api.type.SyncDestination;
 import edu.hawaii.its.api.util.JsonUtil;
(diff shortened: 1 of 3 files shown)


zakgilbert:
Fixing codacy errors rn.

mhodgesatuh (commented):
Given that we have logger output in UpdateTimestampResult, do we need the new logger statements in production?

devgav (approved):
Thanks for finding such a quick fix to this issue. Ik we have have had trouble with the timestamp in the past before as well.

mhodgesatuh (approved):

mhodgesatuh:
For v2.3 we should also have system testing code coverage that ensures the contract to update timestamps for changes to include and exclude lists is fully implemented.

mhodgesatuh:
Julio's recommendation: If the UI allows for more than one person to be sent to those two methods, then you definitely want to set the timestamp after looping through the users.  The timestamp is currently set within the loop.  Zak, can you review and consider making the change as recommended if it makes sense.  It's not clear to me that having the timestamp after the loop will ensure that everything changed gets an updated timestamp though, but then I don't fully understand the code at this point.

PR #391, opened 2022-12-08
Creates predicate class for filtering

KaileeHung:
# Ticket Link

[GROUPINGS-1280](https://uhawaii.atlassian.net/browse/GROUPINGS-1280)

# List of squashed commits

- Starts predicate class for filtering
- Renames PathFilter and finishes tests
- Uses PathFilter in other methods
- Changes so predicate is passed into getGroupPaths
- Adds general formatting fixes
- Makes small fixes to PathFilterTest

# Test Checklist

- [X] Unit Tests Passed:

### DIFF
--- a/src/test/java/edu/hawaii/its/api/service/PathFilterTest.java
+++ b/src/test/java/edu/hawaii/its/api/service/PathFilterTest.java
@@ -1,0 +1,96 @@
+package edu.hawaii.its.api.service;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static edu.hawaii.its.api.service.PathFilter.onlyMembershipPaths;
+import static edu.hawaii.its.api.service.PathFilter.pathHasBasis;
+import static edu.hawaii.its.api.service.PathFilter.pathHasExclude;
+import static edu.hawaii.its.api.service.PathFilter.pathHasInclude;
+import static edu.hawaii.its.api.service.PathFilter.pathHasOwner;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+public class PathFilterTest {
+
+    private String includePath = "bogus:include";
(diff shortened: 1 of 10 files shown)


mhodgesatuh (approved):
I don't see any issues.  The code looks cleaner.

src/test/java/edu/hawaii/its/api/service/PathFilterTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private String excludePath = "bogus:exclude";
+    private String ownerPath = "bogus:owners";
+    private String basisPath = "bogus:basis";
+
+    private List<String> testPathList;
+
+    @Test
+    public void testPathHasInclude() {
+        testPathList = new ArrayList<>();
+        assertEquals(filterPaths(testPathList, pathHasInclude()).size(), 0);
```
zakgilbert, line 29: This is not a big deal, but for the future, the first param of `assertEquals` is the expected value...
```
assertEquals)(int expected, int actual);
Asserts that expected and actual are equal.
```
So if your test fails right now its gonna read...
```
expect: <whatever the value of filterPaths(testPathList, pathHasInclude()).size() is>
actual: 0
```
This can be confusing to other developers and yourself.

src/test/java/edu/hawaii/its/api/service/PathFilterTest.java
```diff
@@ -0,0 +1,95 @@
...
+        testPathList.add(includePath);
+        testPathList.add(includePath);
+        testPathList.add(excludePath);
+        testPathList.add(ownerPath);
+        testPathList.add(basisPath);
+        assertEquals(filterPaths(testPathList, pathHasInclude()).size(), 2);
```
zakgilbert, line 35: This is not a big deal, but for the future, the first param of `assertEquals` is the expected value...
```
assertEquals)(int expected, int actual);
Asserts that expected and actual are equal.
```
So if your test fails right now its gonna read...
```
expect: <whatever the value of filterPaths(testPathList, pathHasInclude()).size() is>
actual: 0
```
This can be confusing to other developers and yourself.

devgav (approved):
thanks for cleaning up our codebase.

PR #395, opened 2022-12-21
Cleaned up membershipResults

carolwong492:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1298)

In conjunction with https://github.com/uhawaii-system-its-ti-iam/uh-groupings-ui/pull/689

# List of squashed commits

- Made small changes

# Test Checklist

- [X] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
@@ -1,26 +1,11 @@
 package edu.hawaii.its.api.controller;
 
+import java.util.List;
+
+import javax.annotation.PostConstruct;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import edu.hawaii.its.api.groupings.GroupingsAddResult;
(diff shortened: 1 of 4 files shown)


mhodgesatuh (approved):
I don't see any issues.

JorWo:
Not sure if I like /filtered-groupings in the URI of the membershipResults endpoint. Maybe for membershipResults it can be `/v2.1/members/{uhIdentifier}/groupings/memberships` and for managePersonResults it can be `/v2.1/members/{uhIdentifier}/groupings/all`.

mhodgesatuh:
We don't have a good name for the expression of basis+include-exclude,
unfortunately, especially given that the expression is the very reason that
the grouping exists in the first place.  Since this expression is how we
derive membership, it does stand to reason that the following would be
appropriate:
 o /v2.1/members/{uhIdentifier}/groupings/memberships
If you read the description of the new entry on the RESTful API page it too
calls out "memberships".

I'm not sure the "all" reference is needed.

JorWo (approved):
Thanks for making the changes to the API URIs.

PR #425, opened 2023-03-30
Implement wrapper for Gc/Wc GetMembers

zakgilbert:
# Ticket Link

[GROUPINGS-1382](https://uhawaii.atlassian.net/browse/GROUPINGS-1382)

# List of squashed commits
- Implement GetMembersCommand.java
- Implement GetMembersResult.java
- Implement  GetMembersResults.java
- Implement GetMembersResultsTest.java
- Implement TestGetMembersCommand.java
- Use new test users from Properties file.
- Use new PropteryLocater Util.

# Test Checklist

- [x] Unit Tests Passed:


mhodgesatuh (commented):
The need for a singular and a plural wrapper class gives me a moments pause.  The plural class in theory should be able to handle 1 or more.  Is the singular wrapper class a convenience for backwards compatibility?  If so the inline comments should explain this since future developers will have the same question most likely.  If there's another reason for the singular class, likewise, inline comments should explain this minor anti pattern.

src/main/java/edu/hawaii/its/api/wrapper/GetMembersResult.java
src/main/java/edu/hawaii/its/api/wrapper/GetMembersResults.java

zakgilbert:
Just added comments.

PR #452, opened 2023-05-25
Updates to Spring Boot 2.7.12, adds Spring Actuator endpoints, removes Thymeleaf

JorWo:
# Ticket Link

[GROUPINGS-1469](https://uhawaii.atlassian.net/browse/GROUPINGS-1469)

# List of squashed commits

- Cleaned up application.properties file
- Updated to Spring Boot 2.7.12
- Removed Thymeleaf, home.html, error.html, HomeController
- Removed /info Spring Actuator endpoint

# Test Checklist

- [x] Unit Tests Passed:


mhodgesatuh (commented):
management.info.os.enabled=true

Unless we have a compelling reason to release this info, I suggest we don't.  Maybe follow up on slack to see if Frank opines otherwise, in which case we would.  Reason is, the more info we release about the service the more info is available for an attempt to compromise the service.

JorWo:
> management.info.os.enabled=true
> 
> Unless we have a compelling reason to release this info, I suggest we don't. Maybe follow up on slack to see if Frank opines otherwise, in which case we would. Reason is, the more info we release about the service the more info is available for an attempt to compromise the service.

Frank has nothing against releasing the info, so I'll remove it.

PR #453, opened 2023-06-05
Standardize the overrides files

FeimeiChen:
# Ticket Link

[Ticket 1144](https://uhawaii.atlassian.net/browse/GROUPINGS-1144)

# List of squashed commits

- Standardize the overrides files

# Test Checklist

- [X] Unit Tests Passed:

### DIFF
--- a/uh-groupings-api-overrides.skeleton.properties
+++ b/uh-groupings-api-overrides.skeleton.properties
@@ -2,26 +2,14 @@
 # Copy this file into that directory and name it uh-groupings-api-overrides.properties.
 # Fill in all blank values.
 
-# UH username
+# Provide your UH username for both of the following.
 groupings.api.localhost.user=
-# UH username
 groupings.api.test.admin_user=
 
+# Grouper client settings. Request the password via filedrop.
 grouperClient.webService.url=https://grouper-test.its.hawaii.edu/grouper-ws/servicesRest/
+grouperClient.webService.login=_groupings_api_2
+grouperClient.webService.password=
 
-# grouper app-user username
-grouperClient.webService.login =
 
-# grouper app-user password
-grouperClient.webService.password =
 
-# Override recipient email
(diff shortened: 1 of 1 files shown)


mhodgesatuh (approved):
Thanks for cleaning this up.

uh-groupings-api-overrides.skeleton.properties
```diff
@@ -1,27 +1,15 @@
...
-# Override recipient email
-email.send.recipient.override=
-
-# Add the following lines below if you want to do a live reload.
-# This is not necessary in order to run the project.
-
-# Path of your uh-groupings-api project see path examples above.
-# project.base-dir=
-# spring.thymeleaf.prefix=${project.base-dir}/src/main/resources/templates/
-# spring.resources.static-locations=${project.base-dir}/src/main/resources/static/
```
michho8, line 27: Was it discussed as a team to get rid of the live reload? It may be appreciated by future team members since it can be a bit annoying to keep rerunning the project after every change.

mhodgesatuh:
Seems to me that the default should be to live reload only in the localhost
environment where a live reload would be helpful, but none of the other
environments, which suggests it belongs in the localhost property file only.

uh-groupings-api-overrides.skeleton.properties
```diff
@@ -1,27 +1,15 @@
-# Create a directory in your home directory called ".<username for your computer>-conf".
+# Create a directory in your home directory called ".<username for your coputer>-conf".
```
JorWo, line 1: Spelling error at "coputer", should be "computer"

JorWo (approved):

uh-groupings-api-overrides.skeleton.properties
FeimeiChen: done, thank you

PR #463, opened 2023-06-20
Move properties out of custom.properties file into application.proper…

FeimeiChen:
# Ticket Link

[Ticket 1487](https://uhawaii.atlassian.net/browse/GROUPINGS-1487)

# List of squashed commits

- Commit 1: Move properties out of custom.properties file into application.properties file
- Commit 2: remove the line

# Test Checklist

- [x] Unit Tests Passed:


fduckart:
Love it! One thing I can't tell, though, is if we happen to have tests to check if this is working as expected? Java Spring tests, I mean.

mhodgesatuh (approved):
Regarding Frank's question about tests; might need more discussion, but it seems that integration testing rather than system testing would be applicable.  Spinning up a separate ticket for the tests is one way of keeping the momentum in this case.

src/main/resources/application.properties
```diff
@@ -19,6 +18,15 @@ cas.saml.tolerance=180000
 cas.send.renew=false
 
 # =========================================================================
+app.iam.request.form=https://www.hawaii.edu/bwiki/display/UHIAM/UH+Groupings+Request+Form
```
JorWo, line 21: We inject this URL from our message.properties file currently, so we can remove this line.

FeimeiChen: done

JorWo (approved):
Looks good, you can use your previously created ticket https://uhawaii.atlassian.net/browse/GROUPINGS-1481 to create tests to check if the application.properties and overrides files have been loaded in.

PR #469, opened 2023-06-22
Sort sync destinations

yertsti:
Add group name to sync dest

# Ticket Link

[1485](https://uhawaii.atlassian.net/browse/GROUPINGS-1485)

# List of squashed commits

- Sort sync destinations
-  Add group name to sync dest

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/groupings/GroupingSyncDestinations.java
+++ b/src/main/java/edu/hawaii/its/api/groupings/GroupingSyncDestinations.java
@@ -3,29 +3,41 @@ package edu.hawaii.its.api.groupings;
 import edu.hawaii.its.api.util.JsonUtil;
 import edu.hawaii.its.api.wrapper.AttributesResult;
 import edu.hawaii.its.api.wrapper.FindAttributesResults;
+import edu.hawaii.its.api.wrapper.Group;
 import edu.hawaii.its.api.wrapper.GroupAttribute;
 import edu.hawaii.its.api.wrapper.GroupAttributeResults;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
+import java.util.regex.Pattern;
 
 /**
  * GroupingSyncDestinations contains the necessary data hydrate the sync destinations tab in a grouping.
  */
 public class GroupingSyncDestinations {
     private String resultCode;
+
     private List<GroupingSyncDestination> syncDestinations;
+
     private List<GroupAttribute> groupAttributes;
 
+    private String groupExtension;
+
+    private final KeyParser keyParser;
+
     public GroupingSyncDestinations(FindAttributesResults findAttributesResults,
-            GroupAttributeResults groupAttributeResults) {
+            GroupAttributeResults groupAttributeResults, KeyParser keyParser) {
+        this.keyParser = keyParser;
+        setGroupExtension(groupAttributeResults.getGroups());
         setGroupAttributes(groupAttributeResults.getGroupAttributes());
         setSyncDestination(findAttributesResults.getResults());
         setResultCode(findAttributesResults.getResultCode());
-
     }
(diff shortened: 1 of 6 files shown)


mhodgesatuh (commented):
Looks like we might need some additional code coverage as a result of this change.

src/main/java/edu/hawaii/its/api/service/GroupingOwnerService.java
```diff
@@ -84,6 +94,24 @@ public GroupingSyncDestinations groupingsSyncDestinations(String currentUser, St
...
+        if (dest.getName().contains("google-group")) {
+            result += GOOGLE_SYNC_DEST_SUFFIX;
+        }
+
+        try {
+            result = pattern.matcher(dest.getDescription()).replaceFirst(result);
+        } catch (PatternSyntaxException e) {
+            result = dest.getDescription();
+        }
+        return result;
```
JorWo, line 115: Could this logic be moved into the GroupingSyncDestinations class? Also see if this bit of code can be refactored to be easier to read, especially the parseKeyVal function.

src/main/java/edu/hawaii/its/api/groupings/GroupingSyncDestinations.java
```diff
@@ -52,10 +72,20 @@ private void setSyncDestination(List<AttributesResult> attributesResults) {
...
     }
+    private void parseKeyVal(String GOOGLE_SYNC_DEST_SUFFIX) {
+        this.syncDestinations.forEach(dest -> {
+            String tmpDescription = dest.getDescription().replaceAll("\\$\\{srhfgs}", groupExtension);
+            if (dest.getName().contains("google-group")) {
+                tmpDescription = tmpDescription + GOOGLE_SYNC_DEST_SUFFIX;
+            }
+            dest.setDescription(tmpDescription);
+        });
+    }
```
JorWo, line 90: I think this parseKeyVal function would be best called in setSyncDestinations rather than the constructor.

JorWo:
Ideally, it would be great if you could import the GOOGLE_SYNC_DEST_SUFFIX from the properties file into GroupingSyncDestinations.java but I know you were running into trouble doing so. Frank might be able to help you.

src/main/java/edu/hawaii/its/api/groupings/GroupingSyncDestinations.java
```diff
@@ -16,13 +18,15 @@ public class GroupingSyncDestinations {
...
+            GroupAttributeResults groupAttributeResults, String GOOGLE_SYNC_DEST_SUFFIX) {
+        setGroupExtension(groupAttributeResults.getGroups());
         setGroupAttributes(groupAttributeResults.getGroupAttributes());
         setSyncDestination(findAttributesResults.getResults());
         setResultCode(findAttributesResults.getResultCode());
-
+        parseKeyVal(GOOGLE_SYNC_DEST_SUFFIX);
     }
 
     public GroupingSyncDestinations() {
```
JorWo, line 32: Also call setGroupExtension in this default constructor.

fduckart:
I think this needs more tweaking.

yertsti:
I tried moving the parseKey function into its own class called KeyParser and passing that in as a parameter in the constructor for the GroupingSyncDestinations. It pretty much just holds the injected values that we need to format the sync destination descriptions. It might be better to just pass those values directly as parameters when constructing GroupingSyncDestinations and making a parseKey method for GroupingSyncDestinations. I don't know.

JorWo:
Refactored the code after talking to Michael about have Grouper hold the "Google Sync Dest Suffix": @gtest.hawaii.edu

https://github.com/uhawaii-system-its-ti-iam/uh-groupings-api/pull/483

PR #478, opened 2023-07-10
Fix missing timestamp update

yertsti:
# Ticket Link

[1500](https://uhawaii.atlassian.net/browse/GROUPINGS-1500)

# List of squashed commits

- Fix missing timestamp update

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/UpdateTimestampService.java
+++ b/src/main/java/edu/hawaii/its/api/service/UpdateTimestampService.java
@@ -8,19 +8,24 @@ import edu.hawaii.its.api.util.JsonUtil;
 import edu.hawaii.its.api.wrapper.UpdateTimestampCommand;
 import edu.hawaii.its.api.wrapper.UpdatedTimestampResult;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 @Service("timestampService")
 public class UpdateTimestampService {
 
     public static final Log logger = LogFactory.getLog(MembershipService.class);
+    @Autowired
+    private GroupPathService groupPathService;
 
     public GroupingsTimestampResult update(GroupingsResult groupingsResult) {
         if (groupingsResult.getResultCode().equals("SUCCESS")) {
+            if (groupPathService.isOwnersGroupPath(groupingsResult.getGroupPath())) {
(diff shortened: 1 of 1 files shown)


mhodgesatuh:
A quick fix for v2.3 is fine since we need production patched ASAP,  but might want to work with Frank to look for a better coding strategy.  Be sure the fix matches the forward-60 seconds requirement evidenced in https://uhawaii.atlassian.net/browse/GROUPINGS-1491.

JorWo:
This PR says it is going into the main branch. Did you mean to merge into devrc-2-3-0?

fduckart:
Roger that. Can we have another Jira to refactor it? Everything about it now looks unorthodox.

mhodgesatuh:
For the record, ticket 1500 is for production, so devrc-2-3-0 branch needs the fix ASAP.

yertsti:
The fix is just quick and dirty and I have no idea how we would test it.

 The builder for updating the timestamp can take multiple groups and it will return an array of WsAssignAttributeResults. Right now the wrapper that holds the results from the command assumes it only gets one result back, so I think if I change everything to a list then we should be able to work for multiple group paths. 

Another way is to pass in a list of group paths to the initial update method and just execute the UpdateTimestampCommand for each path in the list. This way would probably take less work to refactor it.

JorWo (approved):
The dirty fix is fine for now. I would refactor it by modifying the addOwners and removeOwners function in UpdateMembershipService rather than in UpdateTimestampService

JorWo:
If I were to make your current code more readable, I would utilize the GroupPathService and move around the if statements:
```
@Autowired
    private GroupPathService groupPathService;

    public static final Log logger = LogFactory.getLog(MembershipService.class);

    public GroupingsTimestampResult update(GroupingsResult groupingResult) {
        if (groupingResult.getResultCode().equals("SUCCESS")) {
            if (groupPathService.isOwnersGroupPath(groupingResult.getGroupPath())) {
                updateLastModifiedTimestamp(groupPathService.getGroupingPath(groupingResult.getGroupPath()));
            }
            return updateLastModifiedTimestamp(groupingResult.getGroupPath());
        }

        return new GroupingsTimestampResult();
    }
```

PR #484, opened 2023-07-18
Implement API to adhere to 'act-as' requirements.

hokwaichan:
# Ticket Link

[GROUPINGS-1458](https://uhawaii.atlassian.net/browse/GROUPINGS-1458)

# List of squashed commits

- rename assignOwner to setOwner and ownerUsername to ownerUhIdentifier

# Test Checklist

- [x] Unit Tests Passed:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
+++ b/src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
@@ -83,7 +83,7 @@ public class GrouperApiService {
 
     public FindGroupsResults findGroupsResults(String currentUser, String groupPath) {
         FindGroupsResults findGroupsResults = exec.execute(new FindGroupsCommand()
-                .assignOwner(currentUser)
+                .owner(currentUser)
                 .addPath(groupPath));
         return findGroupsResults;
     }
@@ -160,7 +160,7 @@ public class GrouperApiService {
 
     public GroupAttributeResults groupAttributeResults(String currentUser, List<String> attributes, String groupPath) {
         return exec.execute(new GroupAttributeCommand()
-                .assignOwner(currentUser)
+                .owner(currentUser)
                 .addAttributes(attributes)
                 .addGroup(groupPath));
     }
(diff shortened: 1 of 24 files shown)


mhodgesatuh (commented):
Why does removeMembers specify assignOwner, but assignAttributesResults specify setOwner.  What is conveyed by the two being different?  assignGrouperPrivilegesResult is in scope for this question too.

ownerUsername is no longer appropriate since we no longer use “username”. I suspect ownerUid is what is intended here.

Is currentUser meant to be the currentUserUid?  Not sure if it is that or the currentUserUhUuid.

JorWo:
> Why does removeMembers specify assignOwner, but assignAttributesResults specify setOwner. What is conveyed by the two being different? assignGrouperPrivilegesResult is in scope for this question too.
> 
> ownerUsername is no longer appropriate since we no longer use “username”. I suspect ownerUid is what is intended here.
> 
> Is currentUser meant to be the currentUserUid? Not sure if it is that or the currentUserUhUuid.

currentUser is for uhIdentifiers, but currentUser is currently being used in a 100+ places according to my IntelliJ. A ticket should probably be created to address that rather than in this PR.

mhodgesatuh:
if currentUser is equivalent to currentUserUhIdentifier, I think we can
live with just currentUser since the latter seems excessive.

hokwaichan:
> 

Changed name assignOwner to setOwner and  ownerUserName to ownerUhIdentifier

JorWo:
We will need to make sure Admins still have privileges and update the tests for Command classes that have this newly implement "setOwner" function.

JorWo (approved):

JorWo:
I was looking through `GrouperApiService` and came across other methods using `.assignOwner()`. I think we should change all `setOwner` method names to `assignOwner` because the assignOwner/setOwner methods wrap `.assignActAsSubject(subjectLookup(uhIdentifier))`.

mhodgesatuh:
sounds like a good cleanup strategy

fduckart:
I'm a little confused why we need the "assign" part for the setter names, to be honest. Think we should just use "owner". Same thing for "assignActAsSubject".

PR #486, opened 2023-07-27
Update timestamps of multiple group paths

yertsti:
# Ticket Link

[1506](https://uhawaii.atlassian.net/browse/GROUPINGS-1506)

# List of squashed commits

- Update multiple group paths timestamps

# Test Checklist

- [x] Unit Tests Passed:


mhodgesatuh (approved):
I couldn't tell.  Do the unit tests also ensure that a single matching timestamp is used throughout?

yertsti:
Going to try and add some integration tests for the Timestamp service to check that we are actually updating multiple group paths instead of in the unit test using mock data that just happens to have the same timestamps.

src/main/java/edu/hawaii/its/api/service/UpdateMemberService.java
```diff
@@ -350,31 +351,39 @@ private GroupingAddResults addOwners(String groupingPath, List<String> uhIdentif
...
     }
 
     private GroupingAddResult addOwner(String groupingPath, String uhIdentifier) {
         AddMemberResult addMemberResult =
                 grouperApiService.addMember(groupingPath + GroupType.OWNERS.value(), uhIdentifier);
         GroupingAddResult result = new GroupingAddResult(addMemberResult);
-        timestampService.update(result);
+        List<String> groupPaths = new ArrayList<>();
+        groupPaths.add(groupingPath);
+        timestampService.update(result, groupPaths);
```
JorWo, line 366: I think these groupPaths lists created here can be moved into deeper into the timestampService.update() function. I would rename the timestampService.update() function to something like updateAll (you can think of something better) which only takes in the result as a param rather than a result param and list param. Then that new function can handle creating a list of groupPaths to update.

src/main/java/edu/hawaii/its/api/wrapper/UpdateTimestampCommand.java
```diff
@@ -40,6 +59,14 @@ public UpdatedTimestampResult execute() {
...
     }
 
+    public List<UpdatedTimestampResult> execute2() {
+        List<UpdatedTimestampResult> updatedTimestampResults = new ArrayList<>();
+        WsAssignAttributesResults wsAssignAttributesResults = this.gcAssignAttributes.execute();
+        for (int i = 0; i < wsAssignAttributesResults.getWsGroups().length; i++) {
+            updatedTimestampResults.add(new UpdatedTimestampResult(wsAssignAttributesResults, i));
+        }
+        return updatedTimestampResults;
+    }
```
JorWo, line 69: I think it would ideal if we could have the same execute() function that works on both UpdateTimestampCommands created with a groupPath and a list of groupPaths

yertsti: I think if we wanted one "execute" method to handle both a single groupPath and a list of groupPaths, it would probably have to return a list of size equal to the number of paths we want to update. The problem with this is that the UpdateTImestampCommand implements the Command interface and that method signature for execute is that it returns type T. A work around would be make it so it doesn't implement the Command interface, then we could make the execute method return a list or we could make the Command interface's execute method return a List<T>, but everything else that implements it would need to be updated to deal with execute returning a list instead of an object.

src/main/java/edu/hawaii/its/api/service/UpdateTimestampService.java
```diff
@@ -21,18 +27,40 @@ public class UpdateTimestampService {
...
 
+    public List<GroupingTimestampResult> updateBasePath(GroupingResult groupingResult) {
+        if (groupingResult.getResultCode().equals("SUCCESS")) {
+            List<String> groupPaths = new ArrayList<>();
+            groupPaths.add(groupingResult.getGroupPath());
+            groupPaths.add(groupPathService.getGroupingPath(groupingResult.getGroupPath()));
+            return updateLastModifiedTimestamp(groupPaths);
+        }
+        return Collections.emptyList();
+    }
```
JorWo, line 48: I just thought of a better idea. What if we have a single update function but the function is able to parse a groupingPath into a list or a singleton list. For example, if the groupingPath of the groupingResult is tmp:testiwta:testiwta-aux:owners, a list with [tmp:testiwta:testiwta-aux, tmp:testiwta:testiwta-aux:owners] is generated. If the groupingPath of the groupingResult is just tmp:testiwta:testiwta-aux, the list just will be [tmp:testiwta:testiwta-aux]. This functionality should belong in GroupPathService. 

As for the return of the update function, it should be just GroupingTimestampResult rather than a list. I suggest updating the GroupingTimestampResult to hold a list of groupingPaths.

yertsti: I made GroupingTimestampResults and I also needed to make UpdatedTimestampResults. Pretty much everything inside had to be turned into a list. The inherited method getGroupPath() doesn't do anything because it holds a list now instead of just a string.

src/main/java/edu/hawaii/its/api/service/UpdateTimestampService.java
```diff
@@ -21,18 +25,35 @@ public class UpdateTimestampService {
...
+    public GroupPathService groupPathService;
+
+    public GroupingTimestampResults update(GroupingResult groupingResult) {
         if (groupingResult.getResultCode().equals("SUCCESS")) {
-            return updateLastModifiedTimestamp(groupingResult.getGroupPath());
+            List<String> groupPaths = new ArrayList<>();
+            groupPaths.add(groupingResult.getGroupPath());
+            if (groupPathService.isOwnersGroupPath(groupingResult.getGroupPath())) {
+                groupPaths.add(groupPathService.getGroupingPath(groupingResult.getGroupPath()));
+            }
```
JorWo, line 37: I think it would be very nice to put this into a GroupPathService function that returns a list of groupPaths. Also, although our current UHGroupings API does not do this for timestamps, I think a group path of tmp:testiwta:include should update the group paths tmp:testiwta and tmp:testiwta:include. Therefore this update function should not just only cover owners group path.

src/main/java/edu/hawaii/its/api/service/GroupPathService.java
```diff
@@ -112,6 +112,16 @@ private Group getGroup(String groupPath) {
...
 
+    public List<String> getBasePathAndSubPathFromSubPath(String groupPath) {
+        List<String> groupPaths = new ArrayList<>();
+        if (isOwnersGroupPath(groupPath) || isBasisGroupPath(groupPath) || isIncludeGroupPath(groupPath)
+                || isExcludeGroupPath(groupPath)) {
+            groupPaths.add(getGroupingPath(groupPath));
+        }
+        groupPaths.add(groupPath);
+        return groupPaths;
+    }
```
JorWo, line 123: You might be able to use the getExtension and isGroupExtension functions to make this even nicer. For the name of the function, maybe something like extractPaths or getGroupingGroupPaths could work.

src/main/java/edu/hawaii/its/api/service/UpdateTimestampService.java
```diff
@@ -18,21 +22,24 @@
...
 
-    public GroupingTimestampResult update(GroupingResult groupingResult) {
+    @Autowired
+    public GroupPathService groupPathService;
+
+    public GroupingTimestampResults update(GroupingResult groupingResult) {
         if (groupingResult.getResultCode().equals("SUCCESS")) {
-            return updateLastModifiedTimestamp(groupingResult.getGroupPath());
+            List<String> groupPaths = new ArrayList<>();
+            groupPaths.addAll(groupPathService.getBasePathAndSubPathFromSubPath(groupingResult.getGroupPath()));
```
JorWo, line 33: These two lines can be simplified into one line: `List<String> groupPaths = groupPathService.getGroupingGroupPaths(groupingResult.getGroupPath())`

JorWo (approved):
Looks fantastic now, great work! Let me know if you are planning to make any more changes before I merge this.

JorWo:
Actually one more thing, into today's meeting, we should discuss if updating a groupings' composite timestamp is necessary when updating the include or exclude. This is what I found in our [UH Groupings Design 2.x](https://uhawaii.atlassian.net/wiki/spaces/SITARd/pages/12944638/UH+Groupings+Design+2.x#UHGroupingsDesign2.x-Owner%3AAdd%2FDeleteOwnertoGrouping) bwiki: 
image

mhodgesatuh:
And the agenda for today's team meeting

src/main/java/edu/hawaii/its/api/service/UpdateTimestampService.java
```diff
@@ -18,21 +21,23 @@
...
+    public GroupPathService groupPathService;
+
+    public GroupingTimestampResults update(GroupingResult groupingResult) {
         if (groupingResult.getResultCode().equals("SUCCESS")) {
-            return updateLastModifiedTimestamp(groupingResult.getGroupPath());
+            List<String> groupPaths = groupPathService.getGroupingGroupPaths(groupingResult.getGroupPath());
+            return updateLastModifiedTimestamp(groupPaths);
         }
-        return new GroupingTimestampResult();
+        return new GroupingTimestampResults();
```
JorWo, line 34: Sorry for the back and forth changes, it would be better to move the logic out of GroupPathService.getGroupingGroupPaths and back into this update function because updating only the owners group path is specific to timestamps. Also make sure there is test coverage for the logic. Thanks for your hard work on this ticket, this should be the last change to make.

PR #493, opened 2023-09-15
Add endpoint that throws test exception

yertsti:
# Ticket Link

[1517](https://uhawaii.atlassian.net/browse/GROUPINGS-1517)

# List of squashed commits

- Add endpoint that throws test exception

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1ForTesting.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1ForTesting.java
@@ -1,0 +1,26 @@
+package edu.hawaii.its.api.controller;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import edu.hawaii.its.api.exception.ExceptionForTesting;
+
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/api/groupings/v2.1/testing")
+public class GroupingsRestControllerv2_1ForTesting {
+
+    private static final Log logger = LogFactory.getLog(GroupingsRestControllerv2_1ForTesting.class);
+
+    @GetMapping(value = "/exception")
+    @ResponseBody
+    public ResponseEntity throwException() {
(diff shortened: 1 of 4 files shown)


JorWo:
I think we should have a new REST controller just for these sort of endpoints. Hokwai and Jarren's ticket: https://uhawaii.atlassian.net/browse/GROUPINGS-1475 also involves creating an endpoint to test something. Let me know your thoughts on this. We can also have this conversation in Slack.

mhodgesatuh:
Good conversation to move to slack.  I concur that a separate controller
sounds architecturally appropriate.

yertsti:
Not sure if this is what you had in mind, but I made a new REST controller and tossed the throw exception endpoint in there. I did the same on the UI and made /testing/** endpoints require admin role.

JorWo:
> Not sure if this is what you had in mind, but I made a new REST controller and tossed the throw exception endpoint in there. I did the same on the UI and made /testing/** endpoints require admin role.

Looks great! I has me wondering though why our Groupings Rest Controller has v2_1 tagged at the end though. If it is meaningless, I think we should remove it in another ticket.

mhodgesatuh:
v2_1 is there because we transitioned from a v2_0, and may some day
transition to a later version.
My understanding is that this is a fairly common practice.

JorWo:
TestController might be a better name since we do not perform any actions related to Groupings in the controller. Also don't forget to move the tests into their own classes: TestTestController and TestControllerTest. Thanks!

mhodgesatuh:
If we prefix with "Test" it indicates that this is an integration test
class.  What if we called it an AssessmentControllor, or some other synonym
for "test"?

Michael Hodges (he/him/his)
Enterprise Middleware, Identity and Access Management
University of Hawaii, Information Technology Services
***@***.***
808-956-7195 (GMT -10:00)

we>$

JorWo:
I think we can stick with the name GroupingsRestControllerv2_1 then.

JorWo:
I think we can stick with the name GroupingsRestControllerv2_1ForTesting then.

yertsti:
I tossed out all the unused imports and stuff from copying from the original file. I can put them back in if we are going to use them in the future.

JorWo:
> I tossed out all the unused imports and stuff from copying from the original file. I can put them back in if we are going to use them in the future.

That's good, Codacy would label that as an issue.

JorWo:
As for the Codacy issues, we can remove the PropertyLocater because I don't expect us to be creating tests that require reading the grouper.test.properties. For the exception issue, I guess it could be a RuntimeException instead.

JorWo (approved):
Looks great!

PR #494, opened 2023-09-29
Fix update description to use map instead of string.

yertsti:
# Ticket Link

[1535](https://uhawaii.atlassian.net/browse/GROUPINGS-1535)

# List of squashed commits

- Fix update description to use map instead of string.

# Test Checklist


mhodgesatuh (approved):

JorWo:
Is using a map the only option?

yertsti:
I re-read the stack dump and it turns out the error was coming HttpRequestService not the REST controller. The fix for a null body is just adding a null check when building the request. 
I updated the UI PR. I'm going to close this PR.

PR #523, opened 2024-01-24
Update List<GroupingPath> endpoints to return the new GroupingPaths class.

mikhail-shkaralevich:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1614)

# List of squashed commits

- Make optInGroupingPaths return GroupingPaths class.
- Make ownerGroupings return GroupingPaths class.
- Update tests.

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/GroupingAssignmentService.java
+++ b/src/main/java/edu/hawaii/its/api/service/GroupingAssignmentService.java
@@ -101,7 +101,7 @@ public class GroupingAssignmentService {
     /**
      * As a group owner, get a list of grouping paths pertaining to the groups which optInUid can opt into.
      */
-    public List<GroupingPath> optInGroupingPaths(String owner, String optInUid) {
+    public GroupingPaths optInGroupingPaths(String owner, String optInUid) {
         logger.info(String.format("optInGroupingsPaths; owner: %s; optInUid: %s;", owner, optInUid));
 
         List<String> includes = groupingsService.groupPaths(optInUid, pathHasInclude());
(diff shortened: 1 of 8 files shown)


src/main/java/edu/hawaii/its/api/service/GroupingAssignmentService.java
```diff
@@ -111,7 +111,7 @@ public List<GroupingPath> optInGroupingPaths(String owner, String optInUid) {
         optInPaths.removeAll(includes);
         optInPaths = new ArrayList<>(new HashSet<>(optInPaths));
 
-        return groupingsService.getGroupingPaths(optInPaths);
+        return new GroupingPaths(groupingsService.getGroupingPaths(optInPaths));
```
JorWo, line 114: One of Frank's guiding principles when working on the API is to put the functionality as deep as possible. In this case, we should move the functionality to return the `GroupingPaths` class into `groupingsService.getGroupingPaths()`. The groupingsService.getGroupingPaths could look something like this:
```
public List<GroupingPath> getGroupingPaths(List<String> groupPaths) {
    return new GroupingPaths(grouperApiService.findGroupsResults(groupPaths))
}
```
Now you would have add another constructor and setter to `GroupingPaths.java` to that takes in `findGroupsResults`.

mikhail-shkaralevich: We decided to leave as it is, because there is call to another service function, isGroupingPath().

JorWo (approved):

PR #530, opened 2024-02-29
Add EmailRestController and update EmailService for Epic 3.0

JorWo:
# Ticket Link

[GROUPINGS-1637](https://uhawaii.atlassian.net/browse/GROUPINGS-1637)

# List of squashed commits

- Adding missing return type to getAsyncJobResult() and getAnnouncements() in GroupingsRestControllerv2_1
- Cover code to 100% test coverage

# Test Checklist


mhodgesatuh (approved):
It would be cool if the email message format was read into memory from a text file at startup rather than hardcoded into software so that it could be changed more dynamically.

JorWo:
> It would be cool if the email message format was read into memory from a text file at startup rather than hardcoded into software so that it could be changed more dynamically.

That's a good idea, I will implement that.

PR #537, opened 2024-04-01
Implement add operation for member(s) after validating ootb users in OOTB

gitCarrot:
# Ticket Link

[Ticket-1660](https://uhawaii.atlassian.net/browse/GROUPINGS-1660)

# List of squashed commits

- Add groupPath and subjects for ootb users that can be added in groupings and admin

- Move the methods for ootb to OotbGroupingPropertiesService

- Add ootb users that can be added

- Add the logics for separating two cases of subjects : validating current member for hasMemberResults, validating the members who can be added in groupings

- Modify the uhIentifiers with just numbers and username with letters

- Move the subject validation for adding and deleting member by uhIdentifier from subject service to memberAttributeService

- Add the logic for adding and removing the member with not just Uuid, but also usernames.

- Skipping update time stamp in ootp service temporarily

- Modify the mocked ws object for verifying if the data was removed or added

- Change the strategy of validating uhidentifer and username that user wrote before user click the add, remove button

- Convert uh-setting groupPath for ootb users to properties groupPath.

- Use ootb properties service to get the getMemberResult with specific groupPath
- Remove subject operations in wrapper class
- Move isValidOotbUhIdentifier(uhIdentifier) function from memberAttributeService to SubjectService

- Remove unused imports
- Adjust the code style

# Test Checklist


src/main/java/edu/hawaii/its/api/service/MemberAttributeService.java
```diff
@@ -53,7 +56,7 @@ public MemberAttributeResults getMemberAttributeResults(String currentUser, List
             throw new AccessDeniedException();
         }
         List<String> invalidUhIdentifiers = uhIdentifiers.parallelStream()
-                .filter(uhIdentifier -> !subjectService.isValidIdentifier(uhIdentifier))
+                .filter(uhIdentifier -> !subjectService.isValidIdentifier(uhIdentifier) || (grouperService instanceof OotbGrouperApiService && !ootbGroupingPropertiesService.isValidOotbUhIdentifier(uhIdentifier)))
```
gitCarrot, line 59: This code is the only one that added in existed service layer. This code checks whether the uhIdentifier entered by the user is valid when using the OOTB Grouper service. It performs this check when the user clicks the add or remove button with a username or uh number. The code looks up the uhIdentifier in the OOTB user list to verify its validity.        (also I checked that there's no problem on real GROUPER service too)

JorWo: I think you should put this check inside of `subjectService.isValidIdentifier()` instead.

src/main/java/edu/hawaii/its/api/wrapper/GetMembersResults.java
JorWo, line 1: Make sure to remove the unused imports from this class.

src/main/java/edu/hawaii/its/api/service/MemberAttributeService.java
gitCarrot: Originally, the implementation was done inside subjectService.isValidIdentifier(), but later the logic was moved to memberAttributeService. The Subject service, in addition to checking if a user to be added or removed is valid through memberAttribute, also retrieves grouping groups or membership groups with verifying if the current uhIdentifier is valid. Currently, in Ootb system, there is no default user upon login; instead, the current user becomes whoever logs in through cas.its.hawaii.edu, similar to a regular grouper service. However, in this situation where the current user changes with each login, already defined ootb valid users end up being filtered out, resulting in the current user being considered an invalid user. Therefore, the logic has now been separately moved to the memberAttributeService. According to the discussion with Michael today, it's likely that the login through cas.its.hawaii.edu will be removed in the ootb, and logging in as a default user will be set. If that happens, the logic might be moved back to the subject service.

src/main/java/edu/hawaii/its/api/wrapper/GetMembersResults.java
gitCarrot: Alright!

src/main/java/edu/hawaii/its/api/service/MemberAttributeService.java
JorWo: I was thinking of this:
```
public boolean isValidIdentifier(String uhIdentifier) {
    if (grouperService instanceof OotbGrouperApiService) {
        return ootbGroupingPropertiesService.isValidOotbUhIdentifier(uhIdentifier);
    }
    return isValidSubject(getSubject(uhIdentifier));
}
```
Wouldn't this avoid any of the problems you encountered?

gitCarrot: Oh, I just realized the logic that currently in the MemberattributeService was in the isValidSubject, not isValidIdentifier. It still has false value on the isValidOotbUhIdentifier(uhIdentifier) when uhIdentifier is current user uid, but fetching groupings and membership works well since isValidSubject is important logic for validating current user. I will change it with first code that you provided.

JorWo (approved):
Looks good, one last thing, it seems that you don't have the our code style for IntelliJ enabled from here: https://github.com/uhawaii-system-its-ti-iam/uh-intellij-settings. Use the code style and press save on all the files touched by this PR. It's to make sure the imports are in-order and removed if not used.

mhodgesatuh (approved):

JorWo:
Make sure to apply the code style on the /wrapper classes touched by this PR too, thanks!

JorWo (approved):
Looks great!

PR #566, opened 2024-07-28
Implement new API for making request to check if the groupPath is valid.

mikhail-shkaralevich:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1737)

# List of squashed commits

- Create an endpoint to check if path is valid
- Create test for the groupingPathValid() endpoint.

# Test Checklist

# Update API specification:
[Updated API specification](https://uhawaii.atlassian.net/wiki/spaces/SITARd/pages/12944892/RESTful+API+v2.x+Specification#UH-Grouping-API-v2.1)

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
@@ -128,6 +128,17 @@ public class GroupingsRestControllerv2_1 {
     }
 
     /**
+     * Check if a grouping path is valid.
+     */
+    @GetMapping(value = "/grouping/{path:[\\w-:.]+}/is-valid")
+    @ResponseBody
+    public ResponseEntity<Boolean> groupingPathIsValid(@PathVariable String path) {
+        return ResponseEntity
+                .ok()
+                .body(groupingAttributeService.isGroupingPath(path));
+    }
+
+    /**
      * Get a list of all admins.
      */
     @GetMapping(value = "/grouping-admins")
--- a/src/main/java/edu/hawaii/its/api/service/GroupingAttributeService.java
+++ b/src/main/java/edu/hawaii/its/api/service/GroupingAttributeService.java
@@ -16,7 +16,7 @@ import edu.hawaii.its.api.type.GroupingsServiceResult;
 import edu.hawaii.its.api.type.OptRequest;
 import edu.hawaii.its.api.wrapper.AssignAttributesResults;
 import edu.hawaii.its.api.wrapper.AssignGrouperPrivilegesResult;
-import edu.hawaii.its.api.wrapper.GroupAttribute;
+import edu.hawaii.its.api.wrapper.GroupAttributeResults;
 
 @Service
 public class GroupingAttributeService {
@@ -39,6 +39,9 @@ public class GroupingAttributeService {
     @Value("${groupings.api.failure}")
     private String FAILURE;
 
+    @Value("${groupings.api.trio}")
+    private String TRIO;
+
     private static final Log logger = LogFactory.getLog(GroupingAttributeService.class);
 
     private final GrouperService grouperService;
@@ -57,6 +60,13 @@ public class GroupingAttributeService {
         this.memberService = memberService;
         this.groupingsService = groupingsService;
         this.timestampService = timestampService;
+    }
+
+    /**
+    * Checks if a group at path has TRIO attribute.
+    */
+    public boolean isGroupingPath(String path) {
+        return isGroupAttribute(path, TRIO);
     }
 
     /**
@@ -124,10 +134,11 @@ public class GroupingAttributeService {
 
     // Check if attribute is on.
     public boolean isGroupAttribute(String groupPath, String attributeName) {
-        List<GroupAttribute> groupAttributes = grouperService
-                .groupAttributeResults(attributeName, groupPath)
-                .getGroupAttributes();
-        return groupAttributes.stream()
+        GroupAttributeResults groupAttributes = grouperService.groupAttributeResults(attributeName, groupPath);
+        if (groupAttributes == null) {
+            return false;
+        }
+        return groupAttributes.getGroupAttributes().stream()
                 .anyMatch(groupAttribute -> groupAttribute.getAttributeName().equals(attributeName));
     }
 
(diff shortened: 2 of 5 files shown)


mikhail-shkaralevich:
Please note that I used the `GroupPathService.isGroupingPath()` to check if it is a valid path that points to a grouping not a group.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.List;
 
+import edu.hawaii.its.api.service.*;
```
JorWo, line 5: Could you undo this * import? Thanks.

JorWo:
Make sure to add a unit test for GroupingsRestControlerv2_1 as well.

src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
mikhail-shkaralevich: Yeah, not sure why my editor keeps making the same replacement.

mikhail-shkaralevich: Fixed it. Had to change import setting.

src/main/java/edu/hawaii/its/api/service/GroupingAttributeService.java
```diff
@@ -59,6 +62,14 @@ public GroupingAttributeService(GrouperService grouperService,
         this.timestampService = timestampService;
     }
 
+    public boolean isGroupingPath(String path) {
+        try {
+            return isGroupAttribute(path, TRIO);
+        } catch (Exception e) {
+            return false;
+        }
+    }
```
JorWo, line 71: Just curious, was `groupingsService.isGroupingPath()` not sufficient enough?

mikhail-shkaralevich: From what I understood, Eddy and Julio are not sure if the function from `GroupPathService.isGroupingPath()` checks the TRIO attribute. 

The reason why I am using try-catch is because this function returns a server error if the path specified doesn't refer to a grouping nor a group. I can get rid of it, if it is a bad practice.

JorWo: So would it make more sense to call the endpoint isTRIO?

mikhail-shkaralevich: No, you can't have a grouping which is not TRIO.

JorWo: Could you make the following changes? Thanks!:
1. Remove the try-catch:
```
public boolean isGroupingPath(String path) {
    return isGroupAttribute(path, TRIO);
}
```
2. Update `isGroupAttribute` to use null guard:
```
public boolean isGroupAttribute(String groupPath, String attributeName) {
    GroupAttributeResults groupAttributesResults = grouperService.groupAttributeResults(attributeName, groupPath);

    if (groupAttributesResults == null) {
        return false;
    }

    return groupAttributesResults.getGroupAttributes().stream()
            .anyMatch(groupAttribute -> groupAttribute.getAttributeName().equals(attributeName));
}
```
3. Move the code to `GroupPathService.isGroupingPath(String path)`, replacing it entirely since it is not called anywhere
4. Remove `isGroupingPathTest` from `TestGroupingAttributeService` because `TestGroupPathService` already contains a duplicate test.

mikhail-shkaralevich: Good suggestions, done.

mikhail-shkaralevich: The problem with this implementation is that I am getting a beans' dependencies cycle error: 

- GroupPathService needs GroupingAttributeService bean
- GroupingAttributeService needs GroupingsService bean
- GroupingsService needs GroupPathService bean

JorWo: Ok then move the method back to GroupingAttributeService but add a comment above saying that it checks TRIO attribute.

mikhail-shkaralevich: Is using @Lazy or Bean injection through a setter a bad practice?

JorWo: Yeah I would say so. When we still used `@Autowired` above the services, this dependency cycle error would not show up.

src/test/java/edu/hawaii/its/api/service/TestGroupPathService.java
```diff
@@ -74,16 +74,6 @@ public void isValidPath() {
...
 
-    @Test
-    public void isGroupingPath() {
-        assertTrue(groupPathService.isGroupingPath(GROUPING));
-        assertFalse(groupPathService.isGroupingPath(GROUPING_BASIS));
-        assertFalse(groupPathService.isGroupingPath(GROUPING_OWNERS));
-        assertFalse(groupPathService.isGroupingPath(GROUPING_INCLUDE));
-        assertFalse(groupPathService.isGroupingPath(GROUPING_EXCLUDE));
-        assertFalse(groupPathService.isGroupingPath(INVALID_GROUPING_PATH));
-    }
```
JorWo, line 85: Make sure to keep this test in, thanks!

mikhail-shkaralevich:
I wonder, what stops you from merging this?

JorWo:
I put a comment to put back a test you accidentally deleted.

src/test/java/edu/hawaii/its/api/service/TestGroupPathService.java
mikhail-shkaralevich: Should we actually keep this test in, since there is no longer isGroupoingPath(String) function in the grouping path? There is isGroupingPath(Group) function though.

JorWo: Yes, and I just noticed the original isGroupingPath(String) is removed too. I would just keep that for consistency with having a isGroupPath(String). The old isGroupingPath function is still actually accurate because it makes a call to check if the grouping exists in Grouper.

JorWo (approved):
I went ahead and made pushed the changes I requested into the branch for you.

src/test/java/edu/hawaii/its/api/service/TestGroupPathService.java
mikhail-shkaralevich: Someone just told me that the implementation of the old function was not correct, and it shouldn't be used.

mikhail-shkaralevich:
> I went ahead and made pushed the changes I requested into the branch for you.

Thanks!

PR #577, opened 2024-08-14
Call the API to remove membership that does not exist should not trigger a stack dump

gitCarrot:
# Ticket Link

[Ticket-1747](https://uhawaii.atlassian.net/browse/GROUPINGS-1747)

# List of squashed commits

- Add the logic to prevent a race condition in opt-in and out action and avoid calling the grouper API to update the status of someone who is already opted out

- Implement integrations test for checking the case of duplicated opt-in and opt-out commands and modify existed opt-in and opt-out unit tests

- Fix codacy analysis

# Test Checklist


JorWo (approved):
Seems fine.

PR #587, opened 2024-08-28
Fix GrouperPropertyConfigurer and its unit test

mhodgesatuh:
# Ticket Link

https://uhawaii.atlassian.net/browse/GROUPINGS-1678

# List of squashed commits

- n/a

# Test Checklist


src/main/java/edu/hawaii/its/api/configuration/GrouperPropertyConfigurer.java
```diff
@@ -2,51 +2,35 @@
...
 
-    @Autowired
-    public GrouperPropertyConfigurer(Environment env) {
-        this.env = env;
-    }
+    @Value("${grouperClient.webService.login:}")
+    private String webServiceLogin;
+
+    @Value("${grouperClient.webService.password:}")
+    private String webServicePassword;
```
JorWo, line 20: Is the colon ( : ) needed at the end of `@Value`? I know it can be used to define a default value when the value is not found but it seems to define an empty default value.

mhodgesatuh: That was per Frank's recommendation.  I have been consulting with him on this code since Java is not my day job and only consider chatGPT to provide suggestions.  We were trying to prevent the need to use a null guard elsewhere though I confess that I left it in anyway, just to be doubly sure.

JorWo (approved):
Looks great

src/main/java/edu/hawaii/its/api/configuration/GrouperPropertyConfigurer.java
mhodgesatuh: I note that the PR does not have periods at the ends of the sentences of the new comments.  Drat.  Will have to deal with that when the code gets revisited.

JorWo: That's totally fine, I tend to use comments like a bullet point and only use a period for a sentence but not for the very last sentence

PR #596, opened 2024-09-11
Add curated groupings to membership list

imths:
# [Ticket Link](https://uhawaii.atlassian.net/browse/GROUPINGS-257?atlOrigin=eyJpIjoiODdjNTVhN2I2MDJjNDYzYjgxMTI5NTlkZjZjMTk4N2UiLCJwIjoiaiJ9)

[Groupings-257](https://uhawaii.atlassian.net/browse/GROUPINGS-257?atlOrigin=eyJpIjoiODdjNTVhN2I2MDJjNDYzYjgxMTI5NTlkZjZjMTk4N2UiLCJwIjoiaiJ9)

# List of squashed commits

- Add curated groupings to membership list

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/MembershipService.java
+++ b/src/main/java/edu/hawaii/its/api/service/MembershipService.java
@@ -19,8 +19,8 @@ import org.springframework.stereotype.Service;
 
 import edu.hawaii.its.api.exception.AccessDeniedException;
 import edu.hawaii.its.api.exception.UhMemberNotFoundException;
-import edu.hawaii.its.api.groupings.MembershipResults;
 import edu.hawaii.its.api.groupings.ManageSubjectResults;
+import edu.hawaii.its.api.groupings.MembershipResults;
 import edu.hawaii.its.api.type.GroupType;
 import edu.hawaii.its.api.type.ManageSubjectResult;
 import edu.hawaii.its.api.type.MembershipResult;
@@ -76,23 +76,34 @@ public class MembershipService {
         // The disjoint of basis plus include and exclude: (Basis + Include) - Exclude
         List<String> groupingMembershipPaths = disjoint(parentGroupingPaths(basisIncludeExcludePaths),
                 parentGroupingPaths(excludePaths));
+        // A list of all group paths, in which the uhIdentifier is listed (including curated groupings), so we can find the intersection with curated groupings
+        List<String> trioAndCuratedGroupingsPaths = groupingsService.allGroupPaths(uid);
+        // The list of all curated groupings
+        List<String> curatedGroupingsPaths = groupingsService.curatedGroupings();
+        // Intersect the two lists so groupAndCuratedGroupingsPaths is all curated paths the uhIdentifier is listed
+        curatedGroupingsPaths.retainAll(trioAndCuratedGroupingsPaths);
         // Send all the grouping Membership paths to grouper to obtain grouping descriptions.
         List<Group> membershipGroupings = groupPathService.getValidGroupings(groupingMembershipPaths);
         // Get a list of groupings paths of all basis and include groups that have the opt-out attribute.
         List<String> optOutList = groupingsService.optOutEnabledGroupingPaths(parentGroupingPaths(basisAndInclude));
 
-        List<MembershipResult> memberships = createMemberships(membershipGroupings, optOutList);
+        List<MembershipResult> memberships =
+                createMemberships(membershipGroupings, optOutList, curatedGroupingsPaths);
         return new MembershipResults(memberships);
     }
 
-    private List<MembershipResult> createMemberships(List<Group> membershipGroupings, List<String> optOutList) {
+    private List<MembershipResult> createMemberships(List<Group> membershipGroupings, List<String> optOutList,
+            List<String> curatedGroupingsPaths) {
         List<MembershipResult> memberships = new ArrayList<>();
         for (Group grouping : membershipGroupings) {
-            MembershipResult membershipResult = new MembershipResult();
-            membershipResult.setDescription(grouping.getDescription());
-            membershipResult.setPath(grouping.getGroupPath());
-            membershipResult.setName(nameGroupingPath(grouping.getGroupPath()));
-            membershipResult.setOptOutEnabled(optOutList.contains(grouping.getGroupPath()));
+            MembershipResult membershipResult =
+                    new MembershipResult(grouping.getGroupPath(), nameGroupingPath(grouping.getGroupPath()),
+                            grouping.getDescription());
+            membershipResult.setOptOutEnabled(optOutList.contains(membershipResult.getPath()));
+            memberships.add(membershipResult);
+        }
+        for (String grouping : curatedGroupingsPaths) {
+            MembershipResult membershipResult = new MembershipResult(grouping, grouping, "");
             memberships.add(membershipResult);
         }
         return memberships;
--- a/src/main/java/edu/hawaii/its/api/service/GroupingsService.java
+++ b/src/main/java/edu/hawaii/its/api/service/GroupingsService.java
@@ -27,6 +27,9 @@ public class GroupingsService {
     @Value("${groupings.api.trio}")
     private String TRIO;
 
+    @Value("${groupings.api.curated}")
+    private String CURATED;
+
     private final GroupPathService groupPathService;
 
     private final GrouperService grouperService;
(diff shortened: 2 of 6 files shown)


imths:
I have not touched or added anything to the unit or integration tests, and the jacoco reports can be found on the ticket.

mhodgesatuh (commented):
Should allGroupPath be renamed to allTrioGroupingsPaths now that we have curated groupings.  Should curatedGroupings be renamed to allCuratedGroupingsPaths for naming consistency?  I'm assuming the second list contains paths and not just names. If you use different names, they should still be consistent.

hawaii.edu:store : hard coding it here doesn't smell right.  It is either a constant or a property, likely the latter for the most flexibility.

Unit tests?

imths:
> Should allGroupPath be renamed to allTrioGroupingsPaths now that we have curated groupings. Should curatedGroupings be renamed to allCuratedGroupingsPaths for naming consistency? I'm assuming the second list contains paths and not just names. If you use different names, they should still be consistent.
> 
> hawaii.edu:store : hard coding it here doesn't smell right. It is either a constant or a property, likely the latter for the most flexibility.
> 
> Unit tests?

I can go ahead and change the names that you mentioned, and I will try and look for another way of isolating the curated groupings. My idea was that on the [bwiki](https://uhawaii.atlassian.net/wiki/spaces/UHIAM/pages/13402654/uhReleasedGrouping+Values+Available+for+General+Use), all of the curated groupings mentioned shared the same path start. I have updated my [ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-257?atlOrigin=eyJpIjoiMzdhMmRkMGVhYWUzNDQ3NmI3YWVmYzVkNzgzZGUzYjAiLCJwIjoiaiJ9) on the Kanban board to include screenshots of my resulting jacoco reports as well, but have not yet added any testing for this issue.

JorWo:
I would recommend writing a specific Integration test for this.

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -76,6 +76,13 @@ public MembershipResults membershipResults(String currentUser, String uid) {
         // The disjoint of basis plus include and exclude: (Basis + Include) - Exclude
         List<String> groupingMembershipPaths = disjoint(parentGroupingPaths(basisIncludeExcludePaths),
                 parentGroupingPaths(excludePaths));
+        // A list of all group paths, in which the uhIdentifier is listed (including curated groupings), so we can find the intersection with curated groupings
+        List<String> allGroupPaths = groupingsService.allGroupPaths(uid);
+        // The list of all curated groupings
+        List<String> curatedGroupings = groupingsService.curatedGroupings();
+        // Intersect the two lists and add to our original list of groupingMembershipPaths
+        allGroupPaths.retainAll(curatedGroupings);
```
JorWo, line 84: I am wondering if all uhIdentifiers have the same curated groupings? In that case we would not need to perform the intersection at all.

mhodgesatuh: If the question is: if a uhUuid is assigned to one curated grouping, can we assume that it is assigned to all of the curated groupings?  Answer to that question would be no.  A curated grouping is one that we know needs to be released via the uhReleasedGrouping attribute for authorization purposes, but does not belong to a grouping owner.  The IAM team is curating it. The term "curated grouping" should be added to our team dictionary.

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -92,6 +109,11 @@ private List<MembershipResult> createMemberships(List<Group> membershipGroupings
             membershipResult.setDescription(grouping.getDescription());
             membershipResult.setPath(grouping.getGroupPath());
             membershipResult.setName(nameGroupingPath(grouping.getGroupPath()));
+            GroupAttributeResults groupingAttributeResults =
+                    grouperService.groupAttributeResults(CURATED, grouping.getGroupPath());
+            if (groupingAttributeResults.getResultMessage().equals(", Found 1 results.")) {
+                membershipResult.setName(grouping.getGroupPath());
+            }
```
JorWo, line 116: Does checking `groupingAttributeResults.getResultCode().equals("SUCCESS")` here work as well or is the resultCode not sufficient enough compared to checking the message?

JorWo, line 116: Instead of having to make an additional call to Grouper to check which is a curatedGrouping, you could also pass in the list of curatedGroupings as a parameter to this `createMemberships` function. Also I think the MembershipResult object deserves a new constructor where you can pass the name, path, and description into it.

JorWo: Making a call to grouper to check whether every membership is a curated grouping can be slow, which is why I'm suggesting the above. Since you already know which are the curated groupings, you can just pass it into the `createMemberships` function.

mhodgesatuh (commented):
If I understand correctly, you have a specific test to ensure the new code does indeed correctly detect the presence of a curated grouping.  I was specifically looking for this, bravo.

This section still confused me even though I was vocal about the variable names:
        List<String> groupAndCuratedGroupingsPaths = groupingsService.allGroupPaths(uid);
        // The list of all curated groupings
        List<String> curatedGroupingsPaths = groupingsService.curatedGroupings();
        // Intersect the two lists
        
The comment "intersect the two lists" sets me up for the expectation that there is a list of curated groupings only and another for the trio groupings such that the above might read:  allCuratedGroupingsPaths and allTrioGroupingsPaths and then the intersection being allGroupingsPaths.  But the code block begins with groupAndCuratedGroupingsPaths, which doesn't match my expectation of the simple progression of build the two separate lists and then intersecting them.  I must be missing something.

And, yes, I'm suggesting yet another way to name the variables based on the progression that I find intuitive.

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -76,23 +76,34 @@ public MembershipResults membershipResults(String currentUser, String uid) {
         // The disjoint of basis plus include and exclude: (Basis + Include) - Exclude
         List<String> groupingMembershipPaths = disjoint(parentGroupingPaths(basisIncludeExcludePaths),
                 parentGroupingPaths(excludePaths));
+        // A list of all group paths, in which the uhIdentifier is listed (including curated groupings), so we can find the intersection with curated groupings
+        List<String> groupAndCuratedGroupingsPaths = groupingsService.allGroupPaths(uid);
+        // The list of all curated groupings
+        List<String> curatedGroupingsPaths = groupingsService.curatedGroupings();
+        // Intersect the two lists
+        groupAndCuratedGroupingsPaths.retainAll(curatedGroupingsPaths);
```
JorWo, line 84: Based on Michael's comment, the variable names would make sense simplified to a single line:
```
List<String> curatedGroupingsPaths = groupingsService.allGroupPaths(uid).retainAll(groupingsService.curatedGroupings());
```

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -76,23 +76,34 @@ public MembershipResults membershipResults(String currentUser, String uid) {
...
 
-        List<MembershipResult> memberships = createMemberships(membershipGroupings, optOutList);
+        List<MembershipResult> memberships =
+                createMemberships(membershipGroupings, optOutList, groupAndCuratedGroupingsPaths);
         return new MembershipResults(memberships);
     }
 
-    private List<MembershipResult> createMemberships(List<Group> membershipGroupings, List<String> optOutList) {
+    private List<MembershipResult> createMemberships(List<Group> membershipGroupings, List<String> optOutList,
+            List<String> curatedMemberships) {
```
JorWo, line 96: Then `curatedMemberships` should be renamed to `curatedGroupingsPaths` so it's clear what should be passed in.

mhodgesatuh (commented):
groupAndCuratedGroupingsPaths maybe would be better named as trioAndCuratedGroupingsPaths.  Thanks to the ongoing conversation and the constants you introduced in the GroupingService the use of "trio" seems to be the ideal for method names involving trios where curated groupings are also involved.

mhodgesatuh (approved):
Thank you for your patience with the evolution of the method naming suggestions.  Looks good to me.

JorWo (approved):
Looks good!

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -76,23 +76,34 @@ public MembershipResults membershipResults(String currentUser, String uid) {
         // The disjoint of basis plus include and exclude: (Basis + Include) - Exclude
         List<String> groupingMembershipPaths = disjoint(parentGroupingPaths(basisIncludeExcludePaths),
                 parentGroupingPaths(excludePaths));
+        // A list of all group paths, in which the uhIdentifier is listed (including curated groupings), so we can find the intersection with curated groupings
```
fduckart, line 79: Sorry, I missed the window while this PR was still open, but for the future, can you limit each comment to column 80. I will probably result in more comment lines, but that's okay. Also, put a period at the end if you can.  No need to do this for the real code, though.

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -76,23 +76,34 @@ public MembershipResults membershipResults(String currentUser, String uid) {
...
+        List<String> trioAndCuratedGroupingsPaths = groupingsService.allGroupPaths(uid);
+        // The list of all curated groupings
```
fduckart, line 81: This is unnecessary since the code already says that.

PR #597, opened 2024-09-12
Improve ootb user experience and fix minor errors

gitCarrot:
# Ticket Link

[Ticket-1773](https://uhawaii.atlassian.net/browse/GROUPINGS-1773)

# List of squashed commits

- Remove unnecessary data fixtures from data harnass

- Add FindAttributesResults OOTB beans for announcement service

- Enhance JsonUtil to trace the origin class and method of the error and resolve json null safe

- Add properties for OPT_IN and OPT_OUT values in application.properties and refactor the function to manage attribute assignment for updating group path opt-in and opt-out

- Apply the function to get FindAttributesResults from data haness in ootb grouper api service

- Modify the function to update active profiles in multiple times properly and resolve minor unexpected actions in groupings website with ootb.active.user.profiles.json

- Modify the tests for OotbGrouperApiServiceTest and OotbGroupingPropertiesServiceTest

- Change the condition that race condition of opt out happened

- Add the logic to remove AttributeAssign of the OPT_IN and OPT_OUT when user opt-in and opt-out

- Orgnize the object to return in OotbGrouperApiService

- Rewind the changes of JsonUtil and UpdateMemberServiceTest and make separate ticket for the issue

- Add the logic to avoid null point exception when there's no groupings object from json file

- Fix the code with codacy code style
- Mocking hasMembersResult for updateMemberService test
- Add the logic to make difference when adding grouping member between include and basis extension in the json file related to opt in and opt out on the membership page
- Change the name of the hash set that tracks uid that user logged in before
- Remove unnecessary return value and functions

# Test Checklist


src/main/java/edu/hawaii/its/api/service/OotbGroupingPropertiesService.java
```diff
@@ -151,13 +171,28 @@ public void setCurrentUser(String currentUser) {
...
+    }
+
     // Update active user profile with the values from ui request
 
     public OotbActiveProfileResult updateActiveUserProfile(OotbActiveProfile ootbActiveProfile) {
-        List<OotbGrouping> groupings = ootbActiveProfile.getGroupings();
+        if (modificationTracker.contains(ootbActiveProfile.getUid())) {
+            setCurrentUser(ootbActiveProfile.getUid());
+            return null;
+        }
```
JorWo, line 188: Just a question, is the purpose of this to prevent making extra calls below? Since you are using a set, the user's current groupings would not change in this scenario:
1. updateActiveUserProfile gets called with profile 1 and the groupings are set
2. updateActiveUserProfile gets called with profile 2 and the groupings are set
3. updateActiveUserProfile gets called with profile 1 but the groupings do not change

JorWo:
Also make sure to run `git pull --rebase upstream main` because I see some changes from your previous PR.

src/main/java/edu/hawaii/its/api/service/OotbGroupingPropertiesService.java
gitCarrot: Yes, this logic is to prevent extra API call to update active profile after user tried specific active profile at least once before. Once user call API to update specific active profile, those active profile information will be stored in the beans, and the functions in the OotbGroupingPropertiesService will query or update the bean with the value of current user in the <private String currentUser;> in OotbGroupingPropertiesService.

JorWo: Ok that makes more sense. Could you rename `modificationTracker` to something more descriptive like `ootbActiveProfileUuids`?

gitCarrot: Sounds better and clear!

src/main/java/edu/hawaii/its/api/service/OotbGroupingPropertiesService.java
```diff
@@ -151,13 +171,29 @@ public void setCurrentUser(String currentUser) {
...
+
     // Update active user profile with the values from ui request
 
     public OotbActiveProfileResult updateActiveUserProfile(OotbActiveProfile ootbActiveProfile) {
-        List<OotbGrouping> groupings = ootbActiveProfile.getGroupings();
+        // Check the ootb active profile uid that logged in before for caching
+        if (ootbActiveProfileUids.contains(ootbActiveProfile.getUid())) {
+            setCurrentUser(ootbActiveProfile.getUid());
+            return null;
+        }
```
JorWo, line 189: It might be better to return back the OotbActiveProfile in a OotbActiveProfileResult rather than null. We should stray away returning null.

src/main/java/edu/hawaii/its/api/service/UpdateMemberService.java
```diff
@@ -347,6 +352,16 @@ public void checkIfSelfOptOrAdmin(String currentUser, String identifier) {
...
 
+    public boolean hasOptedIn(String groupingPath, String uhIdentifier) {
+        GroupingOptAttributes groupingOptAttributes = groupingOwnerService.groupingOptAttributes(uhIdentifier, groupingPath);
+        return groupingOptAttributes.isOptInOn();
+    }
+
+    public boolean hasOptedOut(String groupingPath, String uhIdentifier) {
+        GroupingOptAttributes groupingOptAttributes = groupingOwnerService.groupingOptAttributes(uhIdentifier, groupingPath);
+        return groupingOptAttributes.isOptOutOn();
+    }
```
JorWo, line 363: These two functions would be better suited inside GroupingOwnerService because they check if a grouping has optIn or optOut enabled, which only a grouping owner should know the status of. Then function names should be isOptInOn and isOptOutOn since the verb "has" implies that a user has recently opted in, which is not what the function does. The function checks if the grouping you own has optIn or optOut enabled. Also an integration test is needed.

mhodgesatuh:
May I suggest: isOptInEnabled and isOptOutEnabled

src/main/java/edu/hawaii/its/api/service/OotbGroupingPropertiesService.java
gitCarrot: Sounds good! Thanks

src/main/java/edu/hawaii/its/api/service/UpdateMemberService.java
gitCarrot: Jordan, sorry for confusion.

JorWo (approved):
Looks good!

src/main/java/edu/hawaii/its/api/service/UpdateMemberService.java
```diff
@@ -45,16 +45,20 @@ public class UpdateMemberService {
...
 
+    private final GroupingOwnerService groupingOwnerService;
+
     public UpdateMemberService(UpdateTimestampService timestampService,
             SubjectService subjectService,
             GroupPathService groupPathService,
             MemberService memberService,
-            GrouperService grouperService) {
+            GrouperService grouperService,
+            GroupingOwnerService groupingOwnerService) {
```
JorWo, line 55: Actually I think you need to remove this too.

JorWo (approved):

PR #610, opened 2024-10-14
Design configuration class for integration of HashiCorp Vault With Spring Boot Application

gitCarrot:
# Ticket Link

[Ticket-1833](https://uhawaii.atlassian.net/browse/GROUPINGS-1833)

# List of squashed commits

- Implement VaultConfig using spring cloud vault
- Fix bean creation error of VaultConfig and completly separate existed local dev environment and docker for running on the both, and modify the tests for vault config class
- Fix code style
- Update the version of dependencies for cloud vault config after checking compatibility with current spring boot version
- Make extra configs to manage the environments that doesn't have active vault on the localhost or dockerhost
- Remove vault core dependency from pom.xml
-  Move token property from source code to docker compose to avoid the situation that
 develoepr push the code with token by mistake, and disable vault in the localhost for the people who don't want to auto create vault bean without installing vault in their os
- Fix property of injection value in the vault config from the secret engine of the vaule

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
+++ b/src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
@@ -1,0 +1,40 @@
+package edu.hawaii.its.api.controller;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.context.annotation.Profile;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import edu.hawaii.its.api.service.OotbGroupingPropertiesService;
+import edu.hawaii.its.api.type.OotbActiveProfile;
+import edu.hawaii.its.api.type.OotbActiveProfileResult;
+
+@RestController
+@Profile("ootb")
+@RequestMapping("/api/groupings/v2.1")
+public class OotbRestController {
+    private static final Log logger = LogFactory.getLog(OotbRestController.class);
+
+    private final OotbGroupingPropertiesService ootbGroupingPropertiesService;
+
+    // Constructor
+    public OotbRestController(OotbGroupingPropertiesService ootbGroupingPropertiesService) {
+        this.ootbGroupingPropertiesService = ootbGroupingPropertiesService;
+    }
+
+     /**
+     * Update Ootb active user profile.
+     */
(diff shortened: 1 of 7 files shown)


mhodgesatuh (commented):
Would like to understand how we might want to use the properties file to configure the URL to the vault.  And how the token for vault access is being provided.

gitCarrot:
@mhodgesatuh I just relocated the Vault token property from the application properties file of the API project to the Docker Compose environment variables. This change ensures that developers do not accidentally commit their tokens to the main branch.

The dependency I incorporated for retrieving Vault secrets is provided by Vault itself. It offers a simpler interface for accessing secrets compared to the original core Vault dependency, which involves numerous complex classes. By simply adding this dependency and configuring a few properties, developers can easily fetch secrets from Vault and directly set them in the Grouper client configuration properties.

One significant advantage of this dependency is that it eliminates the need for maintaining a separate bean for Vault operations. Additionally, I have moved all Vault-related properties into the Docker Compose setup (but in this ticket, I still keep some of the vault properties in the source code ) . I have verified that all these properties function correctly within Docker Compose, enhancing the security of the project.

gitCarrot:
There are ootb controller and controller test that should be contained in the last ticket. 
And I'm trying to figure out rebase error for squashing commits

src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
JorWo, line 1: Huh I thought this was in the API already.

src/main/resources/application-dockerhost.properties
```diff
@@ -2,3 +2,12 @@
...
 spring.config.activate.on-profile=dockerhost
 spring.config.import=optional:file:/overrides/uh-groupings-api-overrides.properties
+
+# =========================================================================
+# Vault Settings
+spring.cloud.vault.uri=https://localhost:8200
+spring.cloud.vault.connection-timeout=5000
+spring.cloud.vault.read-timeout=15000
+
+
```
JorWo, line 12: Make sure to delete these extra spaces.

src/main/java/edu/hawaii/its/api/controller/OotbRestController.java
gitCarrot: I know.. haha

mhodgesatuh (approved):
It is not clear to me why OOTB is also included in this PR.

gitCarrot:
> It is not clear to me why OOTB is also included in this PR.

@mhodgesatuh For some reason, the updates of OOTB that were supposed to be included in the last PR were not applied. Jordan also didn't realize it was not pushed if you see his comment above.

JorWo (approved):

PR #618, opened 2024-11-13
Enhance Flexibility of the IsSoleOwner API Endpoint

gitCarrot:
# Ticket Link

[Ticket-1848](https://uhawaii.atlassian.net/browse/GROUPINGS-1848)

# List of squashed commits

- Enhance Flexibility of the IsSoleOwner API Endpoint
- Remove duplicate logic to check the owner

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/service/GroupingAssignmentService.java
+++ b/src/main/java/edu/hawaii/its/api/service/GroupingAssignmentService.java
@@ -16,9 +16,12 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 import edu.hawaii.its.api.exception.AccessDeniedException;
+import edu.hawaii.its.api.exception.UhMemberNotFoundException;
 import edu.hawaii.its.api.groupings.GroupingGroupMember;
 import edu.hawaii.its.api.groupings.GroupingGroupMembers;
 import edu.hawaii.its.api.groupings.GroupingPaths;
+import edu.hawaii.its.api.type.ApiError;
+import edu.hawaii.its.api.type.ApiValidationError;
 import edu.hawaii.its.api.type.Group;
 import edu.hawaii.its.api.type.GroupType;
 import edu.hawaii.its.api.wrapper.GetMembersResult;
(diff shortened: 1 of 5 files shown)


JorWo:
Is there a situation where getting the number of owners is needed? Otherwise we could change the URI if URI readability is your goal. 

My thoughts are that the isSoleOwner endpoint is straight the point. It returns true or false. With getNumberOfOwners, it is less clear that it could be used to check if a uhIdentifier is a sole owner.

mhodgesatuh:
I'm afraid it was my idea to convert to get number of owners functionality
since refactoring was in progress.  The thought is that it is more flexible
with nearly the same overhead. Wrapping it with a isSoleOwner method would
address your point, to some degree.  My impression is that getting the
number will be useful at some point.  Having said that, I can't totally
dismiss your point, why code for what we don't need at the moment.  And it
is a good one.  If refactoring wasn't happening, we would have never gone
down this path.

Jordan, final call on this is yours.

cheers,
 - Michael

gitCarrot:
I didn't have exact purpose of increasing flexibility instead of readability, but today raine also asked me if there's any function to check the number of the owners to implement group path owners integration test (I still don't know he is going to use it or not ). Maybe it can be used as enhancing integration test or somewhere in the future. Hmm. For now the isSoleOwner function is used only in one function to check if memberToRemove is sole owner or not. Since it is not widely used in the project, do you think it is better to change it? I don't think it cause any vagueness.

src/main/java/edu/hawaii/its/api/service/GroupingAssignmentService.java
```diff
@@ -120,14 +123,17 @@ public GroupingGroupMembers groupingOwners(String currentUser, String groupingPa
...
-                currentUser, groupPath, uidToCheck));
+                currentUser, groupPath, uhIdentifier))  ;
+        if (!memberService.isOwner(uhIdentifier)) {
+            throw new AccessDeniedException();
+        }
         List<GroupingGroupMember> owners = groupingOwners(currentUser, groupPath).getMembers();
-        if (owners.size() >= 2) {
-            return false;
+        if (owners.stream().noneMatch(owner -> owner.getUid().contains(uhIdentifier))) {
+            throw new UhMemberNotFoundException("Owner " + uhIdentifier + " not found");
```
JorWo, line 134: Ok let's go forward with this PR, there may be a use for getting the number of owners in the future. 

This stream check can be removed, the isOwner check above performs the same check.

gitCarrot: Okay! Sounds good

JorWo (approved):

PR #623, opened 2024-12-01
Implement functions for mocking all the grouper results of OOTB tests class with the JSON file for each active profile

gitCarrot:
# Ticket Link

[Ticket-1812](https://uhawaii.atlassian.net/browse/GROUPINGS-1812)

# List of squashed commits

- Add the function in JsonUtil to read json file and convert it to string for mapping to the class, and make the test for it

- Implement the mapping logic for List<OotbActiveProfile> in test configuration class and use it in the OotbGroupingPropertiesServiceTest
- Separate function to map object and list of object from file using existed JsonUtil function
- Add throw error logic when the file is not json or when json file has problem

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/util/JsonUtil.java
+++ b/src/main/java/edu/hawaii/its/api/util/JsonUtil.java
@@ -1,8 +1,13 @@
 package edu.hawaii.its.api.util;
 
+import java.nio.file.Files;
 import java.util.List;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 public class JsonUtil {
@@ -28,7 +33,7 @@ public class JsonUtil {
         try {
             result = new ObjectMapper().readValue(json, type);
         } catch (Exception e) {
-            logger.error("Error: " + type +"; " + e);
+            logger.error("Error: " + type + "; " + e);
         }
         return result;
     }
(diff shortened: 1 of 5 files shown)


src/test/java/edu/hawaii/its/api/configuration/GroupingsTestConfiguration.java
```diff
@@ -52,6 +55,10 @@ private <T> T getWsResultTestData(String propertyName, Class<T> type) {
         return JsonUtil.asObject(json, type);
     }
 
+    public List<OotbActiveProfile> ootbActiveProfilesTestData() {
+        return JsonUtil.asList(JsonUtil.readJsonFileToString("ootb.active.user.profiles.json"), OotbActiveProfile.class);
```
JorWo, line 59: I think it would be preferred to have two new functions asListFromFile and asObjectFromFile instead of readJsonFileToString that will always be wrapped by a asList or asObject function.

islandhouse2000: Yeah sounds good!

gitCarrot: yeap! sounds good

gitCarrot:
Oops another personal token of git in my local machine used for pull.

src/main/java/edu/hawaii/its/api/util/JsonUtil.java
```diff
@@ -44,6 +49,25 @@ public static <T> List<T> asList(final String json, Class<T> type) {
...
+    public static String readJsonFileToString(String filePath) {
+        String result = null;
+        try {
+            Resource resource = new ClassPathResource(filePath);
+            result = Files.readString(resource.getFile().toPath());
+        } catch (Exception e) {
+            logger.error("Error: " + e);
+        }
+        return result;
+    }
```
JorWo, line 61: Let's make this a private function since it would only be used as a helper function in this class.

Also it would be cool if it threw an error if the file was not a json. We can simply just check the file ends in .json.

gitCarrot: okk!! That's great

JorWo (approved):

PR #633, opened 2025-01-22
Add missing removeAdminMembers and removeOwnerships

JorWo:
# Ticket Link

[GROUPINGS-1863](https://uhawaii.atlassian.net/browse/GROUPINGS-1863)

# List of squashed commits

- Add missing removeAdminMember calls to integration tests
- Remove searchGroupingMembers changes

# Test Checklist

### DIFF
--- a/src/test/java/edu/hawaii/its/api/service/TestUpdateMemberService.java
+++ b/src/test/java/edu/hawaii/its/api/service/TestUpdateMemberService.java
@@ -480,14 +480,16 @@ public class TestUpdateMemberService {
             fail("should not throw an exception if user does self opt in when user already opted in");
         }
 
-         try {
+        try {
             updateMemberService.removeIncludeMember(ADMIN, GROUPING, testUid);
             updateMemberService.optIn(ADMIN, GROUPING, testUid);
             assertTrue(memberService.isMember(GROUPING_INCLUDE, testUid));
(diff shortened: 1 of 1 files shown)


mhodgesatuh (approved):
For the following, it is not clear that there is any benefit to the assignment since it is used only once and doesn't improve clarity.

String searchString = "search";
String url = API_BASE_URL + "groupings/" + GROUPING + "/search?searchString=" + searchString;

vs

String url = API_BASE_URL + "groupings/" + GROUPING + "/search?searchString=search";

I approved it regardless.  Your call since there might be something else in play and I'm unaware.

JorWo:
> For the following, it is not clear that there is any benefit to the assignment since it is used only once and doesn't improve clarity.
> 
> String searchString = "search"; String url = API_BASE_URL + "groupings/" + GROUPING + "/search?searchString=" + searchString;
> 
> vs
> 
> String url = API_BASE_URL + "groupings/" + GROUPING + "/search?searchString=search";
> 
> I approved it regardless. Your call since there might be something else in play and I'm unaware.

So the context of this change was the endpoint can break if the user passes in any characters like: `/\[].`. The forward slashes especially would change the URL of the endpoint. Instead of manually cleaning those values out in the UI, this was the better umbrella fix to ensure I did not miss any characters that shouldn't be there.

PR #638, opened 2025-02-26
Refactor RetryExecutorService and ExecutorService into one

mikhail-shkaralevich:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-1914)

# List of squashed commits

- Execute grouper calls in updateOptAttribute with retry option
- Refactor executer services into one with a boolean flag to retry execution

# Test Checklist


mikhail-shkaralevich:
Should I backport this whole change to Groupings 2.5?

mhodgesatuh:
Looking at the original ticket,
https://uhawaii.atlassian.net/browse/GROUPINGS-1914, definitely yes.

src/main/java/edu/hawaii/its/api/service/ExecutorService.java
```diff
@@ -12,15 +12,36 @@
...
+                    logger.debug(text + "execution success");
+                    return result;
+                }
+                if (!retry || i == MAX_RETRIES)
+                    return result;
+            } catch (Exception e) {
+                logger.error(text + e);
+            }
+            if (!retry)
+                return null;
```
JorWo, line 33: There must be a cleaner approach than placing two if statements. Also I think it would be great if the retry boolean was stored in the Command class.

src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
```diff
@@ -367,7 +370,7 @@ public AssignAttributesResults assignAttributesResults(String currentUser, Strin
     public AssignGrouperPrivilegesResult assignGrouperPrivilegesResult(String currentUser, String groupPath,
             String privilegeName,
             String uhIdentifier, boolean isAllowed) {
-        return exec.execute(new AssignGrouperPrivilegesCommand()
+        return exec.execute(RETRY, new AssignGrouperPrivilegesCommand()
                 .owner(currentUser)
                 .setGroupPath(groupPath)
                 .setPrivilege(privilegeName)
```
JorWo, line 376: Being able to set the `retry` boolean as a parameter of this function would add some flexibility rather than hardcoding that it must be retried here.

mikhail-shkaralevich: Do you mean as an optional parameter of this function?

JorWo: Yes, but Java does not easily support optional parameters. So you could just create a second function:
```
public AssignGrouperPrivilegesResult assignGrouperPrivilegesResult(String currentUser, String groupPath, String privilegeName, String uhIdentifier, boolean isAllowed, boolean retry) {
```

src/main/java/edu/hawaii/its/api/service/ExecutorService.java
mikhail-shkaralevich: Did I understand correctly that the idea was to store retry in the command class, so that we could have .retry() option in the command builder chaining?

JorWo: Yes, that is correct.

src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
```diff
@@ -57,7 +57,8 @@ public HasMembersResults hasMembersResults(String currentUser, String groupPath,
         HasMembersResults hasMembersResults = exec.execute(new HasMembersCommand()
                 .owner(currentUser)
                 .assignGroupPath(groupPath)
-                .addUhIdentifiers(uhIdentifiers));
+                .addUhIdentifiers(uhIdentifiers)
+                .setRetry(true));
```
JorWo, line 61: This hasMembersResults does not need a retry, thanks.

JorWo:
Could you check if there is test coverage for the self() methods you added? I believe the Command classes only have an integration test. Thanks.
```
@Override
protected AddMembersCommand self() {
    return this;
}
```

mikhail-shkaralevich:
> Could you check if there is test coverage for the self() methods you added? I believe the Command classes only have an integration test. Thanks.
> 
> ```
> @Override
> protected AddMembersCommand self() {
>     return this;
> }
> ```

I added the test for both unit and integration tests.

src/test/java/edu/hawaii/its/api/wrapper/MockCommand.java
```diff
@@ -0,0 +1,28 @@
...
+    private String executionStatus = null;
+
+    /**
+     * Set the execution status for the command.
+     * Either "SUCCESS" or "FAILURE".
+     */
+    public MockCommand setResultCode(String status) {
+        this.executionStatus = status;
+        return this;
+    }
```
JorWo, line 16: It seems like these are no longer needed.

src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
```diff
@@ -364,7 +364,8 @@ public AssignAttributesResults assignAttributesResults(String currentUser, Strin
                 .setAssignType(assignType)
                 .setAssignOperation(assignOperation)
                 .addGroupPath(groupPath)
-                .addAttribute(attributeName));
+                .addAttribute(attributeName)
+                .setRetry(true));
```
JorWo, line 368: Now with this `.setRetry()`, you should be able to add a `bool retry` parameter to the assignAttributesResults and assignGrouperPrivilegesResult methods.

For example:
```
public AssignAttributesResults assignAttributesResults(String currentUser, String assignType,
            String assignOperation, String groupPath,
            String attributeName, bool retry) {
        return exec.execute(new AssignAttributesCommand()
                .owner(currentUser)
                .setAssignType(assignType)
                .setAssignOperation(assignOperation)
                .addGroupPath(groupPath)
                .addAttribute(attributeName)
                .setRetry(retry));
}
```
NOTE: If this change breaks code in other places, make a copy of this method with `bool retry` rather than updating the current one. It is a bit cleaner that way and you will not need to fix other places of the code.

Then updating line 105 and 118 of GroupingAttributeService.java:
```
// Line 105.
AssignAttributesResults assignAttributesResults = grouperService.assignAttributesResults(
                currentUser, ASSIGN_TYPE_GROUP, assignOperation, groupingPath, attributeName);

// Line 118.
AssignGrouperPrivilegesResult assignGrouperPrivilegesResult = grouperService.assignGrouperPrivilegesResult(currentUser, groupName, privilegeName, EVERY_ENTITY, isSet, true)
```

src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
```diff
@@ -364,7 +364,25 @@ public AssignAttributesResults assignAttributesResults(String currentUser, Strin
...
```
JorWo, line 368: Make sure to remove this `.setRetry(true)`.

src/main/java/edu/hawaii/its/api/service/GrouperApiService.java
```diff
@@ -378,7 +396,25 @@ public AssignGrouperPrivilegesResult assignGrouperPrivilegesResult(String curren
                 .setGroupPath(groupPath)
                 .setPrivilege(privilegeName)
                 .setSubjectLookup(uhIdentifier)
-                .setIsAllowed(isAllowed));
+                .setIsAllowed(isAllowed)
+                .setRetry(true));
```
JorWo, line 400: Same for here, make sure to remove this `.setRetry(true)`.

JorWo (approved):
Looks good, thanks for your hard work on this.

PR #639, opened 2025-02-26
Add owner grouping functionality

lukepag:
# Ticket Link

[1682](https://uhawaii.atlassian.net/browse/GROUPINGS-1682)

# List of squashed commits

- Create all necessary tests
- Create test for getImmediateMembers
- Implement immediate filter logic and create grouping owner object

# Test Checklist

### DIFF
--- a/src/test/java/edu/hawaii/its/api/service/TestGrouperApiService.java
+++ b/src/test/java/edu/hawaii/its/api/service/TestGrouperApiService.java
@@ -1,5 +1,6 @@
 package edu.hawaii.its.api.service;
 
+import static com.jayway.jsonpath.internal.function.ParamType.JSON;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -45,6 +46,8 @@ import edu.hawaii.its.api.wrapper.RemoveMembersResults;
 import edu.hawaii.its.api.wrapper.Subject;
 import edu.hawaii.its.api.wrapper.SubjectsResults;
 
+import edu.internet2.middleware.grouperClient.ws.WsMemberFilter;
+
 @ActiveProfiles("integrationTest")
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 @SpringBootTest(classes = { SpringBootWebApplication.class })
@@ -52,6 +55,12 @@ public class TestGrouperApiService {
 
     @Value("${groupings.api.test.grouping_many}")
     private String GROUPING;
+
+    @Value("${groupings.api.test.grouping_single}")
+    private String GROUPING_SINGLE;
+
+    @Value("${groupings.api.test.grouping_single_owners}")
+    private String GROUPING_SINGLE_OWNERS;
 
     @Value("${groupings.api.test.grouping_many_include}")
     private String GROUPING_INCLUDE;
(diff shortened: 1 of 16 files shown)


src/main/java/edu/hawaii/its/api/groupings/GroupingGroupOwners.java
```diff
@@ -0,0 +1,55 @@
+package edu.hawaii.its.api.groupings;
+
+import edu.hawaii.its.api.wrapper.GetMembersResult;
+import edu.hawaii.its.api.wrapper.SubjectsResults;
+
+/**
+ * When getMembers is called, GroupingGroupOwners holds the information about UH affiliates that are listed in a
+ * group such as include, exclude, owners.
+ */
+public class GroupingGroupOwners implements GroupingResult {
```
JorWo, line 10: I think it makes more sense to call the class "GroupingOwnerMembers" based on the naming of the other class, GroupingGroupMembers. This class you made is specific to the owner group, so "group" should be replaced with "owner" rather than replacing "members".

JorWo:
There are some Codacy issues to fix too.

src/main/java/edu/hawaii/its/api/groupings/MemberResult.java
```diff
@@ -6,6 +6,8 @@ public abstract class MemberResult {
     private String name;
     private String firstName;
     private String lastName;
+    private boolean isOwnerGrouping;
+    private String ownerGroupingPath;
```
JorWo, line 10: Based on your UI code,
```
if(owner.name.includes(':')) {
    owner.isOwnerGrouping = true;
    // Name field in owner groupings initially holds the group path
    owner.ownerGroupingPath = owner.name;
    const splitOwnerPath = owner.name.split(':');
    // Isolate the grouping name from the path
    owner.name = splitOwnerPath[splitOwnerPath.length - 1];
}
```
You can delete the isOwnerGrouping and ownerGroupingPath private variables from this class. Objects in Javascript do not require those keys to be already present to set them.

JorWo: I was looking into if you could extend the MemberResult class with something like OwnerMemberResult, but based on the wrapper implementation, it did not make sense.

src/main/java/edu/hawaii/its/api/groupings/GroupingGroupOwners.java
```diff
@@ -0,0 +1,55 @@
...
+        return resultCode;
+    }
+
+    public void setResultCode(String resultCode) {
+        this.resultCode = resultCode;
+    }
+
+    @Override public String getGroupPath() {
+        return groupPath;
+    }
```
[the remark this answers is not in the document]
lukepag: @JorWo Hey Jordan, I think I left out the overrides because the GroupingResult interface doesn't have any function definitions for setResultCode or setGroupPath. Should I add initializations for them in the GroupingResult object?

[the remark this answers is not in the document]
JorWo: Oh I meant the `@Override` should be above `getResultCode()` and `getGroupPath()`, not to the left. You do not need to put `@Override` for setResultCode or setGroupPath.

JorWo (approved):
Looks good!

src/test/java/edu/hawaii/its/api/service/TestGrouperApiService.java
```diff
@@ -1,5 +1,6 @@
 package edu.hawaii.its.api.service;
 
+import static com.jayway.jsonpath.internal.function.ParamType.JSON;
```
fduckart, line 3: Is this left over?

src/test/java/edu/hawaii/its/api/service/TestGrouperApiService.java
```diff
@@ -45,6 +46,8 @@
 import edu.hawaii.its.api.wrapper.Subject;
 import edu.hawaii.its.api.wrapper.SubjectsResults;
 
+import edu.internet2.middleware.grouperClient.ws.WsMemberFilter;
```
fduckart, line 49: This, too, maybe.

PR #645, opened 2025-04-06
Manage Grouping: Handle not found and access denied errors

hokwaichan:
# Ticket Link

[1943](https://uhawaii.atlassian.net/browse/GROUPINGS-1943)

# List of squashed commits

- Manage Grouping: Handle not found and access denied errors

# Test Checklist

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
+++ b/src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
@@ -663,6 +663,18 @@ public class GroupingsRestControllerv2_1 {
     }
 
     /**
+     * True if user is an owner of the grouping.
+     */
+    @GetMapping(value = "/members/{path:[\\w-:.]+}/{uhIdentifier}/is-owner")
+    @ResponseBody
+    public ResponseEntity<Boolean> hasGroupingOwnerPrivs(@PathVariable String path, @PathVariable String uhIdentifier) {
+        logger.info("Entered REST hasGroupingOwnerPrivs...");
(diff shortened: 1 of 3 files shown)


src/main/java/edu/hawaii/its/api/controller/GroupingsRestControllerv2_1.java
```diff
@@ -662,6 +662,18 @@ public ResponseEntity<Boolean> hasOwnerPrivs(@PathVariable String uhIdentifier)
                 .body(memberService.isOwner(uhIdentifier));
     }
 
+    /**
+     * True if user is an owner of the grouping.
+     */
+    @GetMapping(value = "/members/{path:[\\w-:.]+}/{uhIdentifier}/is-grouping-owner")
```
mikhail-shkaralevich, line 668: I suggest renaming the endpoint to ..../is-owner instead to stay consistent across our naming convention. Also, please update the [API Docs](https://uhawaii.atlassian.net/wiki/spaces/SITARd/pages/12944892/RESTful+API+v2.x+Specification).

hokwaichan: I think we already have is-owner endpoint, do you have other naming suggestions?

mikhail-shkaralevich: We do but it has a different number of path parameters, so there should be no issues with it.

mikhail-shkaralevich:
Looks good, thanks for the update!

mikhail-shkaralevich (approved):

PR #657, opened 2025-07-29
Groupings-2022, Update Endpoint: getNumberOfMemberships

ZhongyuZhang-hi:
# [Groupings-2022, Update Endpoint: getNumberOfMemberships](https://uhawaii.atlassian.net/browse/GROUPINGS-2022)

[Ticket](https://uhawaii.atlassian.net/browse/GROUPINGS-2022)

# List of squashed commits

- Commit 1: Created overloaded MembershipResults method that takes only 1 parameter for getNumberofMemberships, create unit/integration tests for said method, updated getNumberOfMemberships to accept Current User only and use the above methods.
- Commit 2: Removed old MembershipResults method (2 parameters), updated MembershipResults (Controller) to use the new method and to accept current User as the only parameter
- Commit 3: Removed unused variables from updated paths

# Test Checklist

- [X] Exhibits Clear Code Structure:
- [X] Project Unit Tests Passed:
- [X] Project Jasmine Tests Passed:
- [X] Executes Expected Functionality:
- [X] Tested For Incorrect/Unexpected Inputs:

### DIFF
--- a/src/test/java/edu/hawaii/its/api/controller/ErrorControllerAdviceTest.java
+++ b/src/test/java/edu/hawaii/its/api/controller/ErrorControllerAdviceTest.java
@@ -137,7 +137,7 @@ public class ErrorControllerAdviceTest {
         // When current_user and uhIdentifier are the same, but uhIdentifier is not valid
         given(membershipService.membershipResults(uhIdentifier)).willThrow(UhIdentifierNotFoundException.class);
 
-        MvcResult result = mockMvc.perform(get(API_BASE + "/members/memberships", uhIdentifier)
+        MvcResult result = mockMvc.perform(get(API_BASE + "/members/memberships")
                         .header(CURRENT_USER, uhIdentifier))
                 .andExpect(status().isNotFound())
                 .andExpect(jsonPath("$.status").value("NOT_FOUND"))


mikhail-shkaralevich:
Good work on the ticket. Just a few adjustments: instead of overloading `membershipResults` could you just update and leave one instance of it. There's only a single scenario for its use case: to request number of memberships for the current user.

mikhail-shkaralevich:
Also, this ticket involves both UI and API, so make sure that you update the `getNumberOfMemberships()` endpoint in `GroupingsRestController` on the UI as well. And all the related tests of course.

src/main/java/edu/hawaii/its/api/service/MembershipService.java
```diff
@@ -90,6 +90,44 @@ public MembershipResults membershipResults(String currentUser, String uid) {
         return new MembershipResults(memberships);
     }
 
+    /**
+     * Get a list of memberships pertaining to the currentUser. A list of memberships is made up from the groups listings of
+     * (basis + include) - exclude.
+     */
+    public MembershipResults membershipResults(String currentUser) {
```
mikhail-shkaralevich, line 97: Don't overload. Just use this one and remove the other one. We don't need to cluster the codebase with code we don't use.

ZhongyuZhang-hi: The other method is still used by the membershipResults method in the controller, which takes both a current user and an uid as parameters, where the current user is the person trying to access and the uid is the target.

src/test/java/edu/hawaii/its/api/controller/ErrorControllerAdviceTest.java
```diff
@@ -135,16 +135,16 @@ public void testMembershipResultsExceptionHandling() throws Exception {
         String uhIdentifier = "1234";
 
         // When current_user and uhIdentifier are the same, but uhIdentifier is not valid
-        given(membershipService.membershipResults(uhIdentifier, uhIdentifier)).willThrow(UhIdentifierNotFoundException.class);
+        given(membershipService.membershipResults(uhIdentifier)).willThrow(UhIdentifierNotFoundException.class);
 
-        MvcResult result = mockMvc.perform(get(API_BASE + "/members/{uhIdentifier}/memberships", uhIdentifier)
+        MvcResult result = mockMvc.perform(get(API_BASE + "/members/memberships", uhIdentifier)
```
mikhail-shkaralevich, line 140: you don't need the `uhIdentifier `as a parameter here.

PR #696, opened 2026-04-08
Update permissions for CodeQL workflows

rainemendoza:
# Ticket Link

[Ticket](https://uhawaii.atlassian.net/jira/software/c/projects/GROUPINGS/boards/18?quickFilter=75&selectedIssue=GROUPINGS-2167&useStoredSettings=true)

# List of squashed commits

- removed the action permission in job as it is not needed in our workflow due to not having certain artifact-basedlogic or advanced usage
- Commit 2
- Commit 3

# Test Checklist

- [ ] Exhibits Clear Code Structure:
- [ ] Project Unit Tests Passed:
- [ ] Project Integration Tests Passed:
- [ ] Executes Expected Functionality:
- [ ] Tested For Incorrect/Unexpected Inputs:

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -28,9 +28,6 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
-    permissions:
-      actions: read
-
     strategy:
       fail-fast: false
       matrix:


rpirnia (approved):
Looks good!

mhodgesatuh (commented):
ut oh; supply chain attack mitigation requires that we implement permissions, so removing this permission doesn't appear to be a step in the right direction.

rpirnia:
> ut oh; supply chain attack mitigation requires that we implement permissions, so removing this permission doesn't appear to be a step in the right direction.

In accordance with the necessary CodeQL permissions, the code provides `contents: read` and `security-events: write` permissions at the top-level. Raine removed the job-level permissions: `actions: read`, `contents: none` and `security-events: none` since job-level permissions override top-level. 

The permissions are still there (lines 14-16):
```
permissions:
  contents: read
  security-events: write
```
Let me know if that works or if we should make changes!

mhodgesatuh:
ok, that makes sense.  My view during the review didn't make that obvious.  Thanks for the explanation!

rpirnia:
> ok, that makes sense. My view during the review didn't make that obvious. Thanks for the explanation!

No problem!

PR #736, opened 2026-08-13
Add JWT Exception Handling

andrewhe6:
# Ticket Link

[Ticket 2074](https://uhawaii.atlassian.net/browse/GROUPINGS-2074?atlOrigin=eyJpIjoiMjU3MjRiYzVjYzIxNGYyM2IwMzBmYmQyNDIxNmRhNzMiLCJwIjoiaiJ9)

# List of squashed commits

- Add JWT exception handling
- Fix JWT auth filter preventing emails

# Test Checklist

- [ ] Exhibits Clear Code Structure:
- [ ] Project Unit Tests Passed:
- [ ] Project Integration Tests Passed:
- [ ] Executes Expected Functionality:
- [ ] Tested For Incorrect/Unexpected Inputs:

### DIFF
--- a/src/main/java/edu/hawaii/its/api/controller/ErrorControllerAdvice.java
+++ b/src/main/java/edu/hawaii/its/api/controller/ErrorControllerAdvice.java
@@ -20,6 +20,8 @@ import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.context.request.RequestContextHolder;
 import org.springframework.web.context.request.ServletRequestAttributes;
+import io.jsonwebtoken.JwtException;
+import io.jsonwebtoken.security.SignatureException;
 
 import edu.hawaii.its.api.exception.AccessDeniedException;
 import edu.hawaii.its.api.exception.InvalidGroupPathException;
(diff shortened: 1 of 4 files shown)


mhodgesatuh (commented):
Generated with Copilot since you are out of credits:

1) Status code semantics (JwtException → 403)
handleJwtException(JwtException je) currently returns 403 FORBIDDEN for all JWT failures. Can we confirm this matches our API auth contract? For invalid/expired/malformed/missing credentials, 401 UNAUTHORIZED is usually expected, while 403 is typically for authenticated-but-not-authorized cases. If needed, please split by JWT subtype (or map in one place) so status codes are consistent.

2) Stack trace exposure in API response
Both new handlers include stackTrace(ExceptionUtils.getStackTrace(...)) in the ApiError response body. Can we avoid returning full stack traces to clients (at least outside debug/local), since this can expose internal implementation details? Logging/emailing the stack trace is great; response payload should likely be sanitized.

3) Overlapping exception handlers (SignatureException + JwtException)
SignatureException is already a subtype of JwtException, so having separate handlers can drift over time and create inconsistent behavior. Could we consolidate into a single JWT handler with subtype branching (or shared helper) to keep status/message/email policy centralized and easier to maintain?

